PR #4, opened 2019-02-28
Wireless properties, MTU

theri:
This (really short) Pull Request substitutes the Wireless properties with the more abstract data rates based on RFC 8175.
Also, I substituted "available bandwidth" (capacity) with "maximum data rate" in the process, because it's actually the same thing, just available bandwidth (capacty) is supposed to apply to the entire path instead of just one link.
Now this is the theoretical maximum of all links on the path, which might be different from the actually achievable throughput on this path. But I'm not sure we will be able to specify the actually achievable throughput on a path with the definition of path on a per-packet basis and with flows... maybe the flow can have a maximum throughput later? But that's future work.

In this PR, I'm not sure if we need to reference RFC 8175? It inspires our definition, but then again, it's not like they provide an entirely novel thing... And DLEP would be a way to gather these properties, but how to gather properties is out of scope for our draft.

Any opinions?

### DIFF
--- a/draft-enghardt-panrg-path-properties.md
+++ b/draft-enghardt-panrg-path-properties.md
@@ -99,7 +99,7 @@ Disjointness:
 : How disjoint a path is from another path.
 
 Path MTU:
-: The end-to-end maximum transmission unit in one packet.
+: The maximum size, in octets, of an IP packet that can be transmitted without fragmentation on a path segment or path.
 
 Transport Protocols available:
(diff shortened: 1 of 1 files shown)


cyrill-k:
I also like maximum data rate more than "achievable throughput" as the achievable throughput depends on e.g., the protocol and congestion control algorithms used.
I think I would cite RFC8175 to show that our wireless path properties are also defined elsewhere in this way.

theri:
Okay, I added a sentence referencing RFC 8175.
Also, I substituted "end-to-end bandwidth" in the text, because we don't have that anymore, with "end-to-end latency", because we will most likely keep this one.
Does this work?

PR #5, opened 2019-03-05
Terminology

cyrill-k:
Addition of a Terminology section, describing path element, path, flow, aggregated property, measured & potential property.

### DIFF
--- a/draft-enghardt-panrg-path-properties.md
+++ b/draft-enghardt-panrg-path-properties.md
@@ -63,6 +63,35 @@ There are exceptions such as dynamic wireless access properties, but these do no
 This document addresses the first of the questions
 in Path-Aware Networking {{I-D.irtf-panrg-questions}}, which is a product of the PANRG in the IRTF.
 
+# Terminology
+
+This section defines a set of terms used throughout this document. In some cases these terms have been used in other contexts with different meanings so this section attempts to describe each term's meaning with respect to the PANRG activities.
+
+Connection:
+: Connections, as used in this draft, incorporate application, transport, network, data link, and physical layer connections between a set of endpoints. Hereinafter, connections refer to bidirectional or unidirectional connections between two endpoints. How concepts of paths and path properties can be extended to other connection models such as one-to-many or many-to-many connections is not described in this draft.
+
+Path:
+: A path element is a device (including the endpoints), or link used to connect two endpoints and transmit information.
+A path is defined as an ordered set of path elements that can be traversed by a packet at specific points in time.
+For the sake of simplicity, we use the term packet, typically used at the network layer, to describe bit strings transmitted on any layer.
+Depending on which layer the path is looked at, devices, links, and their properties might differ.
+Devices may be hidden, and a set of links may be abstracted into a single link.
+The lifetime of a path overlaps with the lifetime of the corresponding connection but is not restricted to it.
+
+Flow:
+: Several packets traversing the same path elements at specific points in time, can be combined into a flow (e.g., all packets sent within a UDP session).
+As a special case, a flow can consist of just one packet.
+
+Property:
+: A property describes a trait of a set of path elements (e.g., capacity of a link, is device X a firewall, one-way bandwidth which is the minimum of all link bandwidths), or a trait of a flow being sent on a set of path elements (e.g., RTT, one-way delay).
+A property is thus described by a tuple containing the ordered set of path elements, the set of packets traversing the path (the flow) or an empty set if no packets are relevant for the property, the type of trait (e.g., bandwidth), and the value of the trait (e.g., 100mbps).
+
+Aggregated Property:
+: A property can be aggregated over a set of path elements (e.g., loss rate in network backbone calculated using the individual link loss rates), or over a set of packets (e.g., median one-way latency of all packets during the last second), or over both (e.g., average time spent in buffers in the network backbone).
+Aggregation can be numerical (average, sum, min, ...), logical (true if all are true, true if at least X are true, ...), or an arbitrary function which maps a set of input properties to an output property.
+
+Measured & Potential Property:
+: A property can be classified by timescale into a measured property, based on concrete previous and current measurements, and a potential property, which is a property with predicted characteristics, possibly including the reliability of such predictions.
 
 # Domain Properties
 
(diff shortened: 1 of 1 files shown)


draft-enghardt-panrg-path-properties.md
```diff
@@ -63,6 +63,30 @@ There are exceptions such as dynamic wireless access properties, but these do no
...
+A property is thus described by a tuple containing the ordered set of path elements, the set of packets traversing the path (the flow) or an empty set if no packets are relevant for the property, the name of the trait (e.g., maximum data rate), and the value of the trait (e.g., 100mbps).
+
+Aggregated Property:
+: A property can be aggregated over a set of path elements (e.g., MTU in the network backbone as the minimum MTU of the individual path element), or over a set of packets (e.g., median one-way latency of all packets during the last second), or over both (e.g., average time a packets spends in buffers outside the local network).
+Aggregation can be numerical (average, sum, min, ...), logical (true if all are true, true if at least X are true, ...), or an arbitrary function which maps a set of input properties to an output property.
+
+Measured & Potential Property:
+: A property can be classified by timescale into a measured property, based on concrete previous and current measurements, and a potential property, which is a property with predicted characteristics, possibly including the reliability of such predictions.
```
theri: Maybe here an example for a potential property would be useful.
Is this like "expected one-way latency of packets during the next second"?
This would invite the question of how we can even know this. But I would be okay with leaving it in for now and see what PANRG thinks.

draft-enghardt-panrg-path-properties.md
```diff
@@ -63,6 +63,30 @@ There are exceptions such as dynamic wireless access properties, but these do no
...
 
+# Terminology
+
+Path element:
+: A path element is a device (including the endpoints), or link used to connect two endpoints and transmit information on a specific layer.
+Path elements may exist on multiple layers (e.g., the endpoint corresponds to a path element on every layer), may be hidden on higher layers (e.g., a layer 2 switch in the local network), or a path element may be an aggregation of several path elements on a lower layer (e.g., the link connecting the endpoints on the transport layer being an aggregation of all network layer path elements).
+
+Path:
+: A path is defined as an ordered set of path elements at the network layer that can be traversed by a packet at specific points in time.
+For the sake of simplicity, we use the term packet, typically used at the network layer, to describe bit strings transmitted on any layer.
```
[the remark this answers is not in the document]
cyrill-k: I think we can drop it, and if someone complains (e.g., "its not packets, but datagrams that are sent within a UDP session"), then we can either insert it again or fix each inexact use of "packet".

draft-enghardt-panrg-path-properties.md
```diff
@@ -63,6 +63,30 @@ There are exceptions such as dynamic wireless access properties, but these do no
...
 in Path-Aware Networking {{I-D.irtf-panrg-questions}}, which is a product of the PANRG in the IRTF.
 
+# Terminology
+
+Path element:
+: A path element is a device (including the endpoints), or link used to connect two endpoints and transmit information on a specific layer.
+Path elements may exist on multiple layers (e.g., the endpoint corresponds to a path element on every layer), may be hidden on higher layers (e.g., a layer 2 switch in the local network), or a path element may be an aggregation of several path elements on a lower layer (e.g., the link connecting the endpoints on the transport layer being an aggregation of all network layer path elements).
+
+Path:
+: A path is defined as an ordered set of path elements at the network layer that can be traversed by a packet at specific points in time.
```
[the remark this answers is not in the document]
cyrill-k: I guess it does not add any relevant information so we can remove it.

draft-enghardt-panrg-path-properties.md
```diff
@@ -63,6 +63,30 @@ There are exceptions such as dynamic wireless access properties, but these do no
...
+For the sake of simplicity, we use the term packet, typically used at the network layer, to describe bit strings transmitted on any layer.
+
+Flow:
+: Several packets traversing the same path elements at specific points in time, can be combined into a flow (e.g., all packets sent within a UDP session which traverse the same path elements).
```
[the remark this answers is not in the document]
cyrill-k: I would keep the definition on a part of the path instead of on a complete path, in order to be able to for example specify: "median one-way delay outside of local network", which is the median delay of the flow after the second hop to the endpoint.

draft-enghardt-panrg-path-properties.md
cyrill-k: What do you think about the following two examples?

An example of a potential property with a high reliability is the maximum data rate of an ethernet link in the local network during the next day, while a potential property with a lower reliability is the expected one-way latency of packets sent to an endpoint on the other side of the planet during the next second.

theri: Seems like a good starting point. However, now "reliability" is only loosely defined, and I'm not sure if that could be a problem. When I read "reliability", first I was thinking about some statistical error margin or something like this.
Maybe it helps to add another sentence saying what we mean with reliability here? It's an indication of how likely this property is going to be close to the actually measured property, or something like this, right?

draft-enghardt-panrg-path-properties.md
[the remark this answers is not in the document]
theri: Fine with me, it does seem more useful to also have it apply on part of the path.

draft-enghardt-panrg-path-properties.md
cyrill-k: I guess depending on the type of property, the reliability is defined differently, a statistical error margin for numerical values, a probability distribution for categorical values, ...

I guess if we want to have a more formal definition it would be something like (I didn't think thoroughly about this ;):
A potential property *P* is a predicate over a (set of) properties and a time *t* evaluating to true or false depending if some statement holds or does not hold at time *t* (e.g., RTT is between 50 and 100ms in 1 second).
The reliability of a potential property at time *t* is defined as the likelihood of *P* being true.

But I think this would be out of scope of this draft...

theri: You're right, also for non-numerical properties this might be very different than for numerical properties.
I'm okay with leaving it like it is for now and see what the Research Group says.

cyrill-k: Maybe we can just add this sentence and not specify a concrete definition:
The notion of reliability depends on the property, it might be the confidence level and interval for numerical properties or the likelihood that a property holds for non-numerical properties.

theri: Sounds good to me! :)

PR #6, opened 2019-03-06
PAN not only for endpoints. Added three scenarios of systems using PAN.

cyrill-k:
Motivation for why we need the path properties we decided to list.
What do you think about this list?

So far I only added three systems that can use the path properties, but did not specifically say which properties are relevant for which system (e.g., achievable data rate is relevant for p2p file sharing systems). I'm not sure how much into detail we should go here...

### DIFF
--- a/draft-enghardt-panrg-path-properties.md
+++ b/draft-enghardt-panrg-path-properties.md
@@ -51,18 +51,22 @@ Because the current Internet provides an IP-based best-effort bit pipe, endpoint
 A Path Aware Network exposes information about one or multiple paths through the network to endpoints or the network infrastructure.
 
 It is impossible to provide an exhaustive list of path properties, as with every new technology and protocol, novel properties might become relevant.
-In order to specify broadly relevant path properties, scenarios in which knowledge about (additional) path information gives advantages are described here.
+In this document, we specify a set of path properties which might be useful in the following use cases: Traffic policies, network monitoring, and path selection.
 
-Traffic policies leveraging path awareness can be defined.
-Such policies can allow or disallow sending traffic over specific networks or nodes, select an appropriate protocol depending on the capabilities of the on-path devices (e.g., using QUIC instead of TCP if no device blocks UDP traffic), or adjust protocol parameters to the used path.
+- Traffic policies: Network operators or end users can define traffic policies leveraging path awareness.
+Such policies can allow or disallow sending traffic over specific networks or nodes, select an appropriate protocol depending on the capabilities of the on-path devices, or adjust protocol parameters to an existing path.
+An example of a traffic policy is a video streaming application choosing an (initial) video quality based on the achievable data rate, or the monetary cost of the link using a volume-based or flat-rate cost model.
+Another example is an enterprise network where all traffic has to go through a firewall, in which case the endpoint needs to be aware of on-path firewalls.
 
-Network monitoring systems in a path aware network can associate their measurements with the actual path used, and observe if the measurements conform with the advertised path properties.
-Network operators can use the path information to specify comprehensive Quality of Service (QoS) guarantees.
+- Network monitoring: Network operators can use path properties, (e.g., measured by on-path devices), to check whether recent end-user traffic satisfies specific Quality of Service (QoS) requirements, and identify potential problems with their network early on, before the end-user complains.
 
-In the case that an entity can choose between a set of different paths, richer path information can lead to more optimal usage of networked resources.
-An AS can use path information as additional routing metric leading to more informed routing decisions.
-An endpoint can choose an optimal path to a destination if there are several paths, or choose an optimal destination if there are several destinations providing the same service.
-An example of the latter is Application-Layer Traffic Optimization (ALTO), an application layer peer-to-peer protocol allowing endpoints a better-than-random peer selection.
+- Path selection: In some cases, entities can choose to use a certain path (or subset of paths) from a set of paths to achieve a specific goal.
+The possible benefits of a well chosen path varies based on the goal, and thus the only guarantee that a path aware network should give is to not perform worse than the default case without path selection.
+Depending on the goal, paths with different properties are preferred, e.g., retrieving a webpage as quickly as possible requires low latency paths, retrieving a large file in a peer-to-peer network requires paths with high achievable data rate.
+A network (e.g., an AS) can adjust its path selection for internal or external routing based on the path properties.
+In BGP, the Multi Exit Discriminator (MED) attribute decides which path to choose if other attributes are equal; in a path aware network, instead of using this single MED value, other properties such as maximum or available/expected data rate could additionally be used to improve load balancing.
+An endpoint might be able to select between a set of paths, either if there are several paths to the same destination (e.g., if the endpoint is a mobile device with two wireless interfaces, both providing a path), or if there are several destinations, and thus several paths, providing the same service (e.g., Application-Layer Traffic Optimization (ALTO) {{RFC5693}}, an application layer peer-to-peer protocol allowing endpoints a better-than-random peer selection).
+Care needs to be taken when selecting paths based on path properties, as some path properties that depend on the current state of the network can become outdated and not related to path properties of packets sent now.
 
 Such path properties may be relatively dynamic, e.g. current Round Trip Time, close to the origin, e.g. nature of the access technology on the first hop, or far from the origin, e.g. list of ASes traversed.
 
--- a/draft-enghardt-panrg-path-properties.md
+++ b/draft-enghardt-panrg-path-properties.md
@@ -33,6 +33,8 @@ informative:
 
     RFC8175:
 
+    RFC5693:
+
     ANRW18-Metrics: DOI.10.1145/3232755.3232764
 
 --- abstract
@@ -46,8 +48,21 @@ This information is expressed as properties of paths between two endpoints.
 # Introduction
 
 Because the current Internet provides an IP-based best-effort bit pipe, endpoints have little information about paths to other endpoints.
-A Path Aware Network exposes information about one or multiple paths through the network to endpoints,
-so that endpoints can use this information.
+A Path Aware Network exposes information about one or multiple paths through the network to endpoints or the network infrastructure.
+
+It is impossible to provide an exhaustive list of path properties, as with every new technology and protocol, novel properties might become relevant.
+In order to specify broadly relevant path properties, scenarios in which knowledge about (additional) path information gives advantages are described here.
+
+Traffic policies leveraging path awareness can be defined.
+Such policies can allow or disallow sending traffic over specific networks or nodes, select an appropriate protocol depending on the capabilities of the on-path devices (e.g., using QUIC instead of TCP if no device blocks UDP traffic), or adjust protocol parameters to the used path.
+
+Network monitoring systems in a path aware network can associate their measurements with the actual path used, and observe if the measurements conform with the advertised path properties.
+Network operators can use the path information to specify comprehensive Quality of Service (QoS) guarantees.
+
+In the case that an entity can choose between a set of different paths, richer path information can lead to more optimal usage of networked resources.
+An AS can use path information as additional routing metric leading to more informed routing decisions.
+An endpoint can choose an optimal path to a destination if there are several paths, or choose an optimal destination if there are several destinations providing the same service.
+An example of the latter is Application-Layer Traffic Optimization (ALTO), an application layer peer-to-peer protocol allowing endpoints a better-than-random peer selection.
 
 Such path properties may be relatively dynamic, e.g. current Round Trip Time, close to the origin, e.g. nature of the access technology on the first hop, or far from the origin, e.g. list of ASes traversed.
 
--- a/draft-enghardt-panrg-path-properties.md
+++ b/draft-enghardt-panrg-path-properties.md
@@ -53,20 +53,21 @@ A Path Aware Network exposes information about one or multiple paths through the
 It is impossible to provide an exhaustive list of path properties, as with every new technology and protocol, novel properties might become relevant.
 In this document, we specify a set of path properties which might be useful in the following use cases: Traffic policies, network monitoring, and path selection.
 
-- Traffic policies: Network operators or end users can define traffic policies leveraging path awareness.
+- Traffic policies: Entities such as network operators or end users may want to define traffic policies leveraging path awareness.
 Such policies can allow or disallow sending traffic over specific networks or nodes, select an appropriate protocol depending on the capabilities of the on-path devices, or adjust protocol parameters to an existing path.
 An example of a traffic policy is a video streaming application choosing an (initial) video quality based on the achievable data rate, or the monetary cost of the link using a volume-based or flat-rate cost model.
 Another example is an enterprise network where all traffic has to go through a firewall, in which case the endpoint needs to be aware of on-path firewalls.
 
-- Network monitoring: Network operators can use path properties, (e.g., measured by on-path devices), to check whether recent end-user traffic satisfies specific Quality of Service (QoS) requirements, and identify potential problems with their network early on, before the end-user complains.
+- Network monitoring: Network operators can use path properties (e.g., measured by on-path devices), to observe Quality of Service (QoS) characteristics of recent end-user traffic, and identify potential problems with their network early on, before the end-user complains.
 
 - Path selection: In some cases, entities can choose to use a certain path (or subset of paths) from a set of paths to achieve a specific goal.
-The possible benefits of a well chosen path varies based on the goal, and thus the only guarantee that a path aware network should give is to not perform worse than the default case without path selection.
-Depending on the goal, paths with different properties are preferred, e.g., retrieving a webpage as quickly as possible requires low latency paths, retrieving a large file in a peer-to-peer network requires paths with high achievable data rate.
+As the possible benefits of a well chosen path varies based on the goal, as a baseline, a path selection algorithm should aim to not perform worse than the default case most of the time.
+Depending on the goal, an entity may prefer paths with different properties, e.g., retrieving a small webpage as quickly as possible requires low latency paths, or retrieving a large file in a peer-to-peer network requires paths with high achievable data rate.
+Additionally, there may be trade-offs between path properties (e.g., latency and data rate), and entities may influence these trade-offs with their choices.
 A network (e.g., an AS) can adjust its path selection for internal or external routing based on the path properties.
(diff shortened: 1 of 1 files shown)


draft-enghardt-panrg-path-properties.md
```diff
@@ -46,8 +48,21 @@ This information is expressed as properties of paths between two endpoints.
...
 Because the current Internet provides an IP-based best-effort bit pipe, endpoints have little information about paths to other endpoints.
-A Path Aware Network exposes information about one or multiple paths through the network to endpoints,
-so that endpoints can use this information.
+A Path Aware Network exposes information about one or multiple paths through the network to endpoints or the network infrastructure.
+
+It is impossible to provide an exhaustive list of path properties, as with every new technology and protocol, novel properties might become relevant.
+In order to specify broadly relevant path properties, scenarios in which knowledge about (additional) path information gives advantages are described here.
+
+Traffic policies leveraging path awareness can be defined.
+Such policies can allow or disallow sending traffic over specific networks or nodes, select an appropriate protocol depending on the capabilities of the on-path devices (e.g., using QUIC instead of TCP if no device blocks UDP traffic), or adjust protocol parameters to the used path.
```
theri: I'm not quite sure I understand this use case yet, and how it relates to the knowledge about certain path properties.
Maybe it would help to give examples of who might define such traffic policies (network operators? end users?) and what path properties they might use. I think it's better to use active voice than passive voice anyway ("can be defined" --> "network operators or end users can define ...").

In the example given here, I just realized that we have this as a property (whether a certain transport protocol works or not), but it's not like the devices on the path tell us whether they block QUIC -- an endpoint would just have to try a QUIC handshake and fall back to TCP if it doesn't work, right? Then the endpoint could cache the "Transport protocols available" property with its knowledge of whether the endpoint has established a QUIC connection successfully in the past (certain time period), and then the endpoint could be "more confident" in just using QUIC, i.e., it could put less resources into trying out different protocols. But to me, this is not really a policy or a matter of allowing or not allowing traffic, but more, like, an opportunity for performance optimization.

Maybe we can find a more obvious example of an actual policy here.
I'm thinking along the lines of:  An end user might define a policy that they prefer their traffic to pass through a specific network, e.g., their home network rather than a public cellular network. To inform such a policy, it might be useful for the end user to know about properties of the paths such as Monetary Cost or the presence of a specific network on the path (i.e., a fixed-line network where they have a flat-rate contract with the provider, as opposed to a cellular network where the provider charges based on data traffic volume).

Maybe we can also find a policy use case with firewalls or proxies, but I'm not sure if those are easier to understand and agree with for our audience.

draft-enghardt-panrg-path-properties.md
```diff
@@ -46,8 +48,21 @@ This information is expressed as properties of paths between two endpoints.
...
+
+Network monitoring systems in a path aware network can associate their measurements with the actual path used, and observe if the measurements conform with the advertised path properties.
+Network operators can use the path information to specify comprehensive Quality of Service (QoS) guarantees.
```
theri: Here I'm also not quite sure I understand the use case(s).
What does it mean to "associate their measurements with the actual path used"? A monitoring system gets information from on-path devices, e.g., latency measurements, and then it can compare them to... what exactly? Path properties advertised where? In some contract that specifies certain latency properties? Or in some future path-awareness protocol that communicates some network performance to endpoints, and a different party (who?) wants to find out if this information is true?
In the second sentence, do you mean that a network operator could use on-path measurements to, e.g., write somewhere that their 95th quantile latency is x milliseconds, or that their median packet loss is x %? I'm not sure in what specific scenario this would actually happen.

Maybe it's easier to start with something simpler, where we only observe performance and don't promise or advertise anything yet. For example, "Network operators can use path properties, e.g., measured by on-path devices, to check whether recent end-user traffic satisfy specific Quality of Service (QoS) requirements, and identify potential problems with their network early on, before the end-user complains"?
Perhaps here we could cite something later on, like, IPPM or IPFIX or something? I'm not sure I fully understand the context in which these protocols are actually used though...

draft-enghardt-panrg-path-properties.md
```diff
@@ -46,8 +48,21 @@ This information is expressed as properties of paths between two endpoints.
...
+
+In the case that an entity can choose between a set of different paths, richer path information can lead to more optimal usage of networked resources.
+An AS can use path information as additional routing metric leading to more informed routing decisions.
```
theri: Here we mean on an AS level, right? Would that be path properties like "what ASes are further along the path", in a sense of "who is operating them", or something like this? If so, isn't this more about policy and less about efficient resource usage? Or are there other path properties that make sense of an AS level for resource usage, which are not covered by usual BGP attributes (local preference? or was that MED?)?
We should make sure that this use case clearly differs from the policy use case above.

draft-enghardt-panrg-path-properties.md
```diff
@@ -46,8 +48,21 @@ This information is expressed as properties of paths between two endpoints.
...
+An endpoint can choose an optimal path to a destination if there are several paths, or choose an optimal destination if there are several destinations providing the same service.
+An example of the latter is Application-Layer Traffic Optimization (ALTO), an application layer peer-to-peer protocol allowing endpoints a better-than-random peer selection.
```
theri: Optimal in what sense? Less latency? And based on what path properties? In ALTO it was some knowledge of the topology within an AS, or something like that, right?
I'm not sure we have to be that specific, but I would at least like to have a hint of what kind of properties might be useful for this use case.

draft-enghardt-panrg-path-properties.md
```diff
@@ -46,8 +48,21 @@ This information is expressed as properties of paths between two endpoints.
...
+It is impossible to provide an exhaustive list of path properties, as with every new technology and protocol, novel properties might become relevant.
+In order to specify broadly relevant path properties, scenarios in which knowledge about (additional) path information gives advantages are described here.
+
+Traffic policies leveraging path awareness can be defined.
+Such policies can allow or disallow sending traffic over specific networks or nodes, select an appropriate protocol depending on the capabilities of the on-path devices (e.g., using QUIC instead of TCP if no device blocks UDP traffic), or adjust protocol parameters to the used path.
+
+Network monitoring systems in a path aware network can associate their measurements with the actual path used, and observe if the measurements conform with the advertised path properties.
+Network operators can use the path information to specify comprehensive Quality of Service (QoS) guarantees.
+
+In the case that an entity can choose between a set of different paths, richer path information can lead to more optimal usage of networked resources.
```
theri: More optimal usage of networked resources means something like load balancing, right? So the resource is available link capacity?

With path selection, we might also be trying to achieve better (end-to-end) performance, right? I think we should name this aspect here as well, and then provide some examples to illustrate both goals.

theri:
I just went over my notes from IETF 103 again, and I noticed three more points that perhaps make sense to include in the path selection use case:
- Such path selection might hope be to be "better than random".
- Path properties may come with limiltations, e.g., performance-related path properties measured in the past may become outdated and not related to path properties of packets sent now.
- There are trade-offs between different performance-related path properties, e.g., latency and data rate. What constitutes "good performance" also depends on the application's needs.

This is quite a lot, I realize. I can also put in these parts later in a new PR.
I'm just trying to incorporate as many IETF 103 comments as possible before going back and asking the Research Group whether we are on the right track.

draft-enghardt-panrg-path-properties.md
```diff
@@ -46,8 +48,21 @@ This information is expressed as properties of paths between two endpoints.
 # Introduction
 
 Because the current Internet provides an IP-based best-effort bit pipe, endpoints have little information about paths to other endpoints.
-A Path Aware Network exposes information about one or multiple paths through the network to endpoints,
-so that endpoints can use this information.
+A Path Aware Network exposes information about one or multiple paths through the network to endpoints or the network infrastructure.
+
+It is impossible to provide an exhaustive list of path properties, as with every new technology and protocol, novel properties might become relevant.
+In order to specify broadly relevant path properties, scenarios in which knowledge about (additional) path information gives advantages are described here.
```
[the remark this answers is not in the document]
cyrill-k: I don't like the wording of "informing a use case". Maybe we can write it like this:
"In this document, we specify a set of path properties which might be useful in the following use cases: Traffic policies, network monitoring, and path selection."

- Traffic Policies
- ...

draft-enghardt-panrg-path-properties.md
cyrill-k: I'll change to active voice.

Maybe blocking UDP traffic is not a good example. Should we use the following examples (one for each of the use cases):

- An endpoint might require either an on-path or locally installed firewall to send or receive traffic.

- Endpoints communicating with each other within a trusted network might forgo encryption, such as TLS, as long as all devices and links are within the network.

- A video streaming application chooses an (initial) video quality based on the achievable data rate, or the monetary cost of the link, whether path uses a volume-based cost model or a flat-rate model.

Although I'm not too sure about the firewall & encryption example as some people might be hesitant to use path properties for security purposes (is the entity providing path properties trusted? Are path properties authenticated? Are the capabilities of each device authenticated?)...

draft-enghardt-panrg-path-properties.md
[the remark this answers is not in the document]
theri: Sure, that works too!

draft-enghardt-panrg-path-properties.md
theri: All of these examples are about policy, i.e., the first use case, right? 
Yes, I'm not sure about using security examples either, if we do not have anything on trustworthiness path properties yet (which is out of scope for our draft, but still, it would bother me as well). 
However, maybe we could use the first example, e.g., say that in an enterprise network there is a policy that all traffic has to go through a firewall, and then the endpoint wants to be aware of that... about the locally-installed firewall, I'm not sure what it has to do with path properties, as it's all local.
The third example is a good convincing one to me.

I just realized that these examples are also about path selection, just like the name of our third use case, so maybe we have to name the third use case not "path selection" but "traffic optimization" or something similar, to highlight that the third use case is about resource usage and/or performance, not policy.

cyrill-k: Sorry for the confusion. What I meant to say is they are for different aspect of the first use case "traffic policies".

I think "traffic policies" is about path transparency (knowledge about existing paths) rather than path choice (which is handled in the third use case), since we always work with (possibly one) existing path. I would rather keep the title of the third use case as "path selection", as our first use case also has aspects of "traffic optimization".

An example of a traffic policy is a video streaming application choosing an (initial) video quality based on the achievable data rate, or the monetary cost of the link using a volume-based or flat-rate cost model.
Another example is an enterprise network where all traffic has to go through a firewall, in which case the endpoint needs to be aware of on-path firewalls.

draft-enghardt-panrg-path-properties.md
cyrill-k: I agree that it is probably better to make a simple statement that does not require many assumptions. I adjusted the example in the pull request. I think it is fine the way it is now since we only need these use cases to justify our choice of path properties and since most path properties can be relevant for QoS, we should be fine ;)

I'm also not an expert on either of these protocols, but I will take a look at them to see how to justify our properties.

draft-enghardt-panrg-path-properties.md
cyrill-k: I guess for AS level path selection, the difference between path selection and traffic policy is very small, as the way ASes setup their policies is by selecting paths...

I guess for ASes (assuming they use BGP), "traffic policies" mean allowing and disallowing certain AS paths based on information from path properties about these ASes; and "path selection" means choosing the paths with the most benefits (e.g., lowest latency, highest maximum data rate, best load balancing, ...).
I think in BGP, regarding performance, we can only argue about intra-network links and links incoming and outgoing of the network, since we have little information about the rest of the path (list of AS numbers only).

draft-enghardt-panrg-path-properties.md
theri: Yep, I think we can use these examples like that.
And yes, I agree to leave path selection separate from "traffic policies".

draft-enghardt-panrg-path-properties.md
cyrill-k: Removed statement about "optimal" paths. Added a sentence that paths with different properties might be preferred in the beginning of this issue.

draft-enghardt-panrg-path-properties.md
theri: Cool, thanks!

draft-enghardt-panrg-path-properties.md
cyrill-k: Example for AS path selection: Load balancing
General example for path selection: reduce webpage load time & retrieving large file in p2p network.

draft-enghardt-panrg-path-properties.md
```diff
@@ -46,8 +48,25 @@ This information is expressed as properties of paths between two endpoints.
...
+A network (e.g., an AS) can adjust its path selection for internal or external routing based on the path properties.
+In BGP, the Multi Exit Discriminator (MED) attribute decides which path to choose if other attributes are equal; in a path aware network, instead of using this single MED value, other properties such as maximum or available/expected data rate could additionally be used to improve load balancing.
+An endpoint might be able to select between a set of paths, either if there are several paths to the same destination (e.g., if the endpoint is a mobile device with two wireless interfaces, both providing a path), or if there are several destinations, and thus several paths, providing the same service (e.g., Application-Layer Traffic Optimization (ALTO) {{RFC5693}}, an application layer peer-to-peer protocol allowing endpoints a better-than-random peer selection).
+Care needs to be taken when selecting paths based on path properties, as some path properties that depend on the current state of the network can become outdated and not related to path properties of packets sent now.
```
theri: This still reads a bit confusing to me, because we have "the current state of the network" and "packets sent now", which sounds like we are talking about the same point in time.
Suggestion: "[…] as path properties that were recently measured may have become outdated and, thus, useless to predict the path properties of packets sent now."

draft-enghardt-panrg-path-properties.md
```diff
@@ -46,8 +48,25 @@ This information is expressed as properties of paths between two endpoints.
...
 
 Because the current Internet provides an IP-based best-effort bit pipe, endpoints have little information about paths to other endpoints.
-A Path Aware Network exposes information about one or multiple paths through the network to endpoints,
-so that endpoints can use this information.
+A Path Aware Network exposes information about one or multiple paths through the network to endpoints or the network infrastructure.
+
+It is impossible to provide an exhaustive list of path properties, as with every new technology and protocol, novel properties might become relevant.
+In this document, we specify a set of path properties which might be useful in the following use cases: Traffic policies, network monitoring, and path selection.
+
+- Traffic policies: Network operators or end users can define traffic policies leveraging path awareness.
```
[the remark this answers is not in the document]
cyrill-k: Agreed & changed.

draft-enghardt-panrg-path-properties.md
```diff
@@ -46,8 +48,25 @@ This information is expressed as properties of paths between two endpoints.
...
+Such policies can allow or disallow sending traffic over specific networks or nodes, select an appropriate protocol depending on the capabilities of the on-path devices, or adjust protocol parameters to an existing path.
+An example of a traffic policy is a video streaming application choosing an (initial) video quality based on the achievable data rate, or the monetary cost of the link using a volume-based or flat-rate cost model.
+Another example is an enterprise network where all traffic has to go through a firewall, in which case the endpoint needs to be aware of on-path firewalls.
+
+- Network monitoring: Network operators can use path properties, (e.g., measured by on-path devices), to check whether recent end-user traffic satisfies specific Quality of Service (QoS) requirements, and identify potential problems with their network early on, before the end-user complains.
```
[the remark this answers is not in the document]
cyrill-k: Yes, that sounds better & is more general.

draft-enghardt-panrg-path-properties.md
```diff
@@ -46,8 +48,25 @@ This information is expressed as properties of paths between two endpoints.
...
+
+- Path selection: In some cases, entities can choose to use a certain path (or subset of paths) from a set of paths to achieve a specific goal.
+The possible benefits of a well chosen path varies based on the goal, and thus the only guarantee that a path aware network should give is to not perform worse than the default case without path selection.
```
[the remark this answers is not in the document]
cyrill-k: You are right, this property is not achievable in some cases.
What about this sentence?

As the possible benefits of a well chosen path varies based on the goal, as a baseline, a path selection algorithm should aim to not perform worse than the default case most of the time.

draft-enghardt-panrg-path-properties.md
```diff
@@ -46,8 +48,25 @@ This information is expressed as properties of paths between two endpoints.
...
+Depending on the goal, paths with different properties are preferred, e.g., retrieving a webpage as quickly as possible requires low latency paths, retrieving a large file in a peer-to-peer network requires paths with high achievable data rate.
```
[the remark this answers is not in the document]
cyrill-k: Did you mean "Additionally, there may be trade-offs between path properties, e.g., latency and data rate, and ~~path elements~~ **entities** may influence these trade-offs with their choices."?

Or am I misunderstanding what is meant by trade-offs?
My understanding is: Entity selecting optimal path w.r.t. property A may need to make compromises on property B.

draft-enghardt-panrg-path-properties.md
cyrill-k: Changed sentence to:
[...], as path properties that were previously measured may have become outdated and thus useless to predict the path properties of packets sent now.

draft-enghardt-panrg-path-properties.md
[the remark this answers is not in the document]
theri: Yep, sounds good.

draft-enghardt-panrg-path-properties.md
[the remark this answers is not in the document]
theri: I meant that, for example, if I choose a path because it's low latency and then transmit a lot of data over it, it might no longer be low latency.
Also I'm thinking about what Gorry said at the IETF 103 session: On the "Dynamic properties" slide, he commented that "you can trade any of these properties for another", and something about modems. To be honest, I'm not sure I fully understand what he was saying and why you can trade all these properties for each other... but that's where I'm trying to go.
As it's not entirely clear, I'm okay with leaving it out for now and see if we get a better understanding of these trade-offs.

draft-enghardt-panrg-path-properties.md
theri: Yes, that works! My professor is very strict about always putting commas around the "thus", but I guess we don't have to be so strict here ;)

PR #8, opened 2019-03-07
Redefine domain, backbone, and dynamic properties

theri:
I went over our property and category definitions and adjusted them to our new terminology.
In the process, PvDs went out of the draft, because they are related to how to get domain properties, not what domain properties exist. But maybe if we find we are missing PvD, we can add it again later.

Also, I realized that several properties relate to both path elements in the same domain and in the backbone. Overall, these categories are quite endpoint-centric, and maybe we want to rethink them at some point - not now though. ;)

### DIFF
--- a/draft-enghardt-panrg-path-properties.md
+++ b/draft-enghardt-panrg-path-properties.md
@@ -29,8 +29,6 @@ informative:
 
     I-D.irtf-panrg-questions:
 
-    RFC7556:
-
     RFC8175:
 
     ANRW18-Metrics: DOI.10.1145/3232755.3232764
@@ -66,63 +64,62 @@ in Path-Aware Networking {{I-D.irtf-panrg-questions}}, which is a product of the
 
 # Domain Properties
 
-Domain path properties usually relate to the access network within the first hop or the first few hops.
-Endpoints can influence domain properties for example by switching from a WiFi to a cellular interface, changing their data plan to increase throughput, or moving closer to a wireless access point which increases the signal strength.
+Domain path properties relate to path elements within the same Autonomous System (AS), thus, in the same administrative domain as an endpoint considering them.
 
-A large amount of information about domain properties exists.
-Properties related to configuration can be queried using provisioning domains (PvDs).
-A PvD is a consistent set of network configuration information as defined in {{RFC7556}}, e.g., relating to a local network interface.
-This may include source IP address prefixes, IP addresses of DNS servers, name of an HTTP proxy server, DNS suffixes associated with the network, or default gateway IP address.
-As one PvD is not restricted to one local network interface, a PvD may also apply to multiple paths.
+Due to the potential physical proximity and pre-existing trust or contractual relationships between endpoints and path elements within the same administrative domain, domain properties may be more easily available to the endpoint than the properties of path elements outside of its administrative domain.
 
-Access Technology present on the path:
-: The lower layer technology on the first hop, for example, WiFi, Wired Ethernet, or Cellular. This can also be more detailed, e.g., further specifying the Cellular as 2G, 3G, 4G, or 5G technology, or the WiFi as 802.11a, b, g, n, or ac. These are just examples, this list is not exhaustive, and there is no common index of identifiers here. Note that access technologies further along the path may also be relevant, e.g., a cellular backbone is not only the first hop, and there may be a DSL line behind the WiFi.
+Furthermore, endpoints may be able to influence both which domain they are in and which path elements in this domain to connect to, and they may be able to influence the properties of path elements within this domain.
+For example, a user might select between multiple potential path elements which are directly adjacent by selecting between multiple available WiFi Access Points. When connected to an Access Point, they may move closer to a it to enable their device to use a different access technology, potentially increasing the data rate available to the device.
+Another example is a user changing their data plan to reduce the Monetary Cost to transmit a given amount of data across a network.
+
+
+Access Technology:
+: The physical or link layer technology used on one or multiple path elements for transmitting a flow. The Access Technology may be given in an abstract way, e.g., as a WiFi, Wired Ethernet, or Cellular link, or as a specific technology, e.g., as a 2G, 3G, 4G, or 5G cellular link, or an 802.11a, b, g, n, or ac WiFi link. Other path elements relevant to the access technology may include on-path devices, such as elements of a cellular backbone network. Note that there is no common registry of possible values for this property.
 
 Monetary Cost:
-: This is information related to billing, data caps, etc. It could be the allowed monthly data cap, the start and end of a billing period, the monetary cost per Megabyte sent or received, etc.
+: The price to be paid to transmit a specific flow across a set of path elements.
+
+Presence of a certain network function on the path:
+: Indicates that a certain path element performs a certain network function on a flow, e.g., whether the path element acts as a proxy, as a firewall, or performs Network Address Translation (NAT). This path element may be either in the same administrative domain as the endpoint or in a different administrative domain, i.e., the backbone.
 
 
 # Backbone Properties
 
-Backbone path properties relate to non-dynamic path properties that are not within the endpoint's domain.
-They are likely to stay constant within the lifetime of a connection, since Internet "backbone" routes change infrequently.
-These properties usually change on the timescale of seconds, minutes, or hours, when the route changes.
+Backbone path properties relate to path elements that are in a different administrative domain than an endpoint considering them, thus, in the backbone from the endpoint's point of view.
 
-Even if these properties change, endpoints can neither specify which backbone nodes to use, nor verify data was sent over these nodes.
-An endpoint can for example choose its access provider, but cannot choose the backbone path to a given destination since the access provider will make their own policy-based routing decision.
+Typically, backbone properties are less easily available to an endpoint than domain properties, due to the potential increased distance and the lack of pre-existing trust or contractual relationship.
 
-Presence of certain device on the path:
-: Could be the presence of a certain kind of middlebox, e.g., a proxy, a firewall, a NAT.
+Additionally, endpoints are less likely to be able to influence which path elements form their path in the backbone, as well as their properties.
 
-Presence of a packet forwarding node or specific Autonomous System on a path:
-: Indicates that traffic goes through a certain node or AS, which might be relevant for deciding the level of trust this path provides.
+Some path properties relate to the entire path, part of which often lies outside of an endpoint's administrative domain. Thus, such properties are listed as Backbone Properties.
+
+
+Administrative Entity:
+: The administrative entity, e.g., the AS, to which a path element or set of path elements belongs.
 
 Disjointness:
-: How disjoint a path is from another path.
+: For a set of two paths, the set of path elements which are shared.
 
 Path MTU:
 : The maximum size, in octets, of an IP packet that can be transmitted without fragmentation on a path segment or path.
 
 Transport Protocols available:
-: Whether a specific transport protocol can be used to establish a connection over this path. An endpoint may know this because it has cached whether it could successfully establish, e.g., a QUIC connection, or an MPTCP subflow.
+: Whether a specific transport protocol can be used to establish a connection over a path or path segment (ordered set of path elements). An endpoint may cache its knowledge about recent successfully established connections using specific protocols, e.g., a QUIC connection, or an MPTCP subflow, over a specific path.
 
 Protocol Features available:
-: Whether a specific feature within a protocol is known to work over this path, e.g., ECN, or TCP Fast Open.
+: Whether a specific protocol feature is available over this path, e.g., Explicit Congestion Notification (ECN), or TCP Fast Open.
 
 
 # Dynamic Properties
 
-Dynamic Path Properties are expected to change on the timescale of milliseconds.
-They usually relate to the state of the path, such as the current end-to-end latency.
-Some of these properties may depend only on the first hop or on the access network, some may depend on the entire path.
-Properties related to a single layer 2 domain are abstracted from the used physical and link layer technology, similar to {{RFC8175}}.
+Dynamic path properties relate to a path or path element with respect to the transmission of an individual packet or of a flow over a path element or path.
+Properties related to a path element which constitutes single layer 2 domain are abstracted from the used physical and link layer technology, similar to {{RFC8175}}.
 
 Typically, Dynamic Properties can only be approximated and sampled, and might be made available in an aggregated form, such as averages or minimums.
 Dynamic Path Properties can be measured by the endpoint itself or somethere in the network.
 See {{ANRW18-Metrics}} for a discussion of how to measure some dynamic path properties at the endpoint.
 
-
-These properties may be symmetric or asymmetric. For example, an asymmetric property may be different in the upstream direction and in the downstream direction from the point of view of a particular host.
+Some dynamic properties are defined in different directions for the same path element, e.g., for transmitting and receiving packets.
 
 
 Maximum Data Rate (Transmit/Receive):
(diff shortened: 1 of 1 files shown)


draft-enghardt-panrg-path-properties.md
```diff
@@ -66,63 +64,62 @@ in Path-Aware Networking {{I-D.irtf-panrg-questions}}, which is a product of the
...
+Presence of a certain network function on the path:
+: Indicates that a certain path element performs a certain network function on a flow, e.g., whether the path element acts as a proxy, as a firewall, or performs Network Address Translation (NAT). This path element may be either in the same administrative domain as the endpoint or in a different administrative domain, i.e., the backbone.
 
 
 # Backbone Properties
 
-Backbone path properties relate to non-dynamic path properties that are not within the endpoint's domain.
-They are likely to stay constant within the lifetime of a connection, since Internet "backbone" routes change infrequently.
-These properties usually change on the timescale of seconds, minutes, or hours, when the route changes.
+Backbone path properties relate to path elements that are in a different administrative domain than an endpoint considering them, thus, in the backbone from the endpoint's point of view.
```
cyrill-k: I would change this sentence to conform with the modified Domain property definition:
"Backbone path properties relate to path elements that are not within the endpoint's domain, and thus in the backbone from the endpoint's point of view."

draft-enghardt-panrg-path-properties.md
```diff
@@ -66,63 +64,62 @@ in Path-Aware Networking {{I-D.irtf-panrg-questions}}, which is a product of the
...
 
-Even if these properties change, endpoints can neither specify which backbone nodes to use, nor verify data was sent over these nodes.
-An endpoint can for example choose its access provider, but cannot choose the backbone path to a given destination since the access provider will make their own policy-based routing decision.
+Typically, backbone properties are less easily available to an endpoint than domain properties, due to the potential increased distance and the lack of pre-existing trust or contractual relationship.
```
cyrill-k: I would change ~~easily available~~ to **accessible**

draft-enghardt-panrg-path-properties.md
```diff
@@ -66,63 +64,62 @@ in Path-Aware Networking {{I-D.irtf-panrg-questions}}, which is a product of the
...
 
-Presence of certain device on the path:
-: Could be the presence of a certain kind of middlebox, e.g., a proxy, a firewall, a NAT.
+Additionally, endpoints are less likely to be able to influence which path elements form their path in the backbone, as well as their properties.
 
-Presence of a packet forwarding node or specific Autonomous System on a path:
-: Indicates that traffic goes through a certain node or AS, which might be relevant for deciding the level of trust this path provides.
+Some path properties relate to the entire path, part of which often lies outside of an endpoint's administrative domain. Thus, such properties are listed as Backbone Properties.
```
cyrill-k: Some path properties relate to the entire path, part of which often lies outside of an endpoint's ~~administrative~~ domain. Thus, such properties are listed as Backbone Properties.

draft-enghardt-panrg-path-properties.md
```diff
@@ -66,63 +64,62 @@ in Path-Aware Networking {{I-D.irtf-panrg-questions}}, which is a product of the
...
+
+
+Administrative Entity:
+: The administrative entity, e.g., the AS, to which a path element or set of path elements belongs.
 
 Disjointness:
-: How disjoint a path is from another path.
+: For a set of two paths, the set of path elements which are shared.
```
cyrill-k: I would extend the description of disjointness a bit:

"For a set of two paths, the number of shared path elements can be a measure of intersection (e.g., Jaccard coefficient, which is the number of shared elements divided by the total number of elements), and conversely, the number of non-shared path elements can be a measure of disjointness (e.g., 1 - Jaccard coefficient).
A multipath protocol might use disjointness of paths as a metric to reduce the number of single points of failure."

draft-enghardt-panrg-path-properties.md
```diff
@@ -66,63 +64,62 @@ in Path-Aware Networking {{I-D.irtf-panrg-questions}}, which is a product of the
...
+: Whether a specific protocol feature is available over this path, e.g., Explicit Congestion Notification (ECN), or TCP Fast Open.
 
 
 # Dynamic Properties
 
-Dynamic Path Properties are expected to change on the timescale of milliseconds.
-They usually relate to the state of the path, such as the current end-to-end latency.
-Some of these properties may depend only on the first hop or on the access network, some may depend on the entire path.
-Properties related to a single layer 2 domain are abstracted from the used physical and link layer technology, similar to {{RFC8175}}.
+Dynamic path properties relate to a path or path element with respect to the transmission of an individual packet or of a flow over a path element or path.
```
cyrill-k: I would use the previously defined path segment ;)
"Dynamic path properties relate to a path segment with respect to the transmission of an individual packet or flow over this path segment."

Maybe we should define path segment in the terminology section?

draft-enghardt-panrg-path-properties.md
```diff
@@ -66,63 +64,62 @@ in Path-Aware Networking {{I-D.irtf-panrg-questions}}, which is a product of the
 
 # Domain Properties
 
-Domain path properties usually relate to the access network within the first hop or the first few hops.
-Endpoints can influence domain properties for example by switching from a WiFi to a cellular interface, changing their data plan to increase throughput, or moving closer to a wireless access point which increases the signal strength.
+Domain path properties relate to path elements within the same Autonomous System (AS), thus, in the same administrative domain as an endpoint considering them.
```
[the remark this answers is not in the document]
theri: I wanted a clear definition, but you're right, "domain" is not necessarily "AS". 
Can we drop the "usually" and just say that it's the same administrative domain? In your example, that administrative domain would be ETH, not "SWITCH", I'd say. But now I'm wondering if there's a formal definition for administrative domain...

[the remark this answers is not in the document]
theri: I wanted a clear definition, but you're right, "domain" is not necessarily "AS". 
Now I'm wondering if there's a formal definition for administrative domain...

draft-enghardt-panrg-path-properties.md
```diff
@@ -66,63 +64,62 @@ in Path-Aware Networking {{I-D.irtf-panrg-questions}}, which is a product of the
...
+
+Due to the potential physical proximity and pre-existing trust or contractual relationships between endpoints and path elements within the same administrative domain, domain properties may be more easily available to the endpoint than the properties of path elements outside of its administrative domain.
```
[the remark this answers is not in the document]
theri: Okay, sounds good.

draft-enghardt-panrg-path-properties.md
```diff
@@ -66,63 +64,62 @@ in Path-Aware Networking {{I-D.irtf-panrg-questions}}, which is a product of the
...
 
-A large amount of information about domain properties exists.
-Properties related to configuration can be queried using provisioning domains (PvDs).
-A PvD is a consistent set of network configuration information as defined in {{RFC7556}}, e.g., relating to a local network interface.
-This may include source IP address prefixes, IP addresses of DNS servers, name of an HTTP proxy server, DNS suffixes associated with the network, or default gateway IP address.
-As one PvD is not restricted to one local network interface, a PvD may also apply to multiple paths.
+Furthermore, endpoints may be able to influence both which domain they are in and which path elements in this domain to connect to, and they may be able to influence the properties of path elements within this domain.
+For example, a user might select between multiple potential path elements which are directly adjacent by selecting between multiple available WiFi Access Points. When connected to an Access Point, they may move closer to a it to enable their device to use a different access technology, potentially increasing the data rate available to the device.
```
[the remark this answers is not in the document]
theri: Agreed, thanks!

draft-enghardt-panrg-path-properties.md
```diff
@@ -66,63 +64,62 @@ in Path-Aware Networking {{I-D.irtf-panrg-questions}}, which is a product of the
...
 
-Access Technology present on the path:
-: The lower layer technology on the first hop, for example, WiFi, Wired Ethernet, or Cellular. This can also be more detailed, e.g., further specifying the Cellular as 2G, 3G, 4G, or 5G technology, or the WiFi as 802.11a, b, g, n, or ac. These are just examples, this list is not exhaustive, and there is no common index of identifiers here. Note that access technologies further along the path may also be relevant, e.g., a cellular backbone is not only the first hop, and there may be a DSL line behind the WiFi.
+
+Access Technology:
+: The physical or link layer technology used on one or multiple path elements for transmitting a flow. The Access Technology may be given in an abstract way, e.g., as a WiFi, Wired Ethernet, or Cellular link, or as a specific technology, e.g., as a 2G, 3G, 4G, or 5G cellular link, or an 802.11a, b, g, n, or ac WiFi link. Other path elements relevant to the access technology may include on-path devices, such as elements of a cellular backbone network. Note that there is no common registry of possible values for this property.
```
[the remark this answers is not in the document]
theri: I think for an endpoint this only makes sense within the same domain, as the endpoint might be able to influence that technology. In the backbone, the endpoint doesn't care if some link is 1G or 10G Ethernet, if there's Fiber involved or not, etc., does it? At most it cares about congestion on this link and the resulting latency or data rate bottleneck. 

Then I guess I'll change first sentence of the definition to only include path elements in the same domain. I don't think that we need to add a further definition to this section, but perhaps we need a definition for "domain" in our Terminology section.

draft-enghardt-panrg-path-properties.md
theri: Well, I don't say AS here, but I can change "within a different administrative domain" to "not within the same domain", then it's more clearly separated from domain properties.

draft-enghardt-panrg-path-properties.md
theri: Fine with me.

draft-enghardt-panrg-path-properties.md
theri: Agreed.

Now I realize that moving the network functions from the Backbone properties to the domain properties was probably a mistake, as this property also relates to the entire path. So I will put them back in the Backbone Properties section.

draft-enghardt-panrg-path-properties.md
theri: Nice. I didn't know about this and it seems pretty useful. Thanks!

draft-enghardt-panrg-path-properties.md
theri: Yes, we definitely should!

As the terminology section is merged now, I'm merging it into this branch and then adding the missing definitions.

theri: Oh, and I just realized that as is, a path segment is the same as a path. But a path should be defined end-to-end, right? So I'm adding "between two endpoints" to the path definition.

draft-enghardt-panrg-path-properties.md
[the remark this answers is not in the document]
cyrill-k: I had the same thoughts when I looked at our definitions. Administrative Domain is defined in RFC1136 and the relation to an AS is: an administrative domain consisting of a single routing domain is analogous to an AS.
But an AS cannot span multiple administrative domains, which means our local network (which is an example of our definition of domain) cannot be an administrative domain.

draft-enghardt-panrg-path-properties.md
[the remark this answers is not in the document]
cyrill-k: That's true.

I'm not sure if our definition of domain in the Terminology section is not confusing since we mention both administrative control and administrative domain which do not relate to the same concept.

draft-enghardt-panrg-path-properties.md
cyrill-k: Yes that sounds good

cyrill-k:
Maybe the easiest solution regarding the definition of "Domain" is to not provide a separate definition in the Terminology section, and just keep the definition of "within the first few hops", as it is given now in the "Domain Property" section. At least for the submission tomorrow, and then we can ask the group if they have an idea how to clearly define it.

draft-enghardt-panrg-path-properties.md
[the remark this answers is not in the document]
theri: I see, thanks for looking it up.
Even though it's defined like this elsewhere, I think we could still have a different definition in our draft. But I'll remove the definition from the Terminology section for now, as it's not clear enough and maybe confusing. Hopefully we can find a better definition, maybe with a little help from the Research Group.

draft-enghardt-panrg-path-properties.md
```diff
@@ -91,63 +92,62 @@ The notion of reliability depends on the property, it might be the confidence le
...
 
-Access Technology present on the path:
-: The lower layer technology on the first hop, for example, WiFi, Wired Ethernet, or Cellular. This can also be more detailed, e.g., further specifying the Cellular as 2G, 3G, 4G, or 5G technology, or the WiFi as 802.11a, b, g, n, or ac. These are just examples, this list is not exhaustive, and there is no common index of identifiers here. Note that access technologies further along the path may also be relevant, e.g., a cellular backbone is not only the first hop, and there may be a DSL line behind the WiFi.
+Access Technology:
+: The physical or link layer technology used for transmitting or receiving a flow on one or multiple path elements in the same administative domain. The Access Technology may be given in an abstract way, e.g., as a WiFi, Wired Ethernet, or Cellular link, or as a specific technology, e.g., as a 2G, 3G, 4G, or 5G cellular link, or an 802.11a, b, g, n, or ac WiFi link. Other path elements relevant to the access technology may include on-path devices, such as elements of a cellular backbone network. Note that there is no common registry of possible values for this property.
```
cyrill-k: The physical or link layer technology used for transmitting or receiving a flow on one or multiple path elements in the same domain.

draft-enghardt-panrg-path-properties.md
```diff
@@ -91,63 +92,62 @@ The notion of reliability depends on the property, it might be the confidence le
...
-: This is information related to billing, data caps, etc. It could be the allowed monthly data cap, the start and end of a billing period, the monetary cost per Megabyte sent or received, etc.
+: The price to be paid to transmit a specific flow across a set of path elements.
 
 
 # Backbone Properties
 
-Backbone path properties relate to non-dynamic path properties that are not within the endpoint's domain.
-They are likely to stay constant within the lifetime of a connection, since Internet "backbone" routes change infrequently.
-These properties usually change on the timescale of seconds, minutes, or hours, when the route changes.
+Backbone path properties relate to path elements that not within the same domain as an endpoint considering them, thus, in the backbone from the endpoint's point of view.
```
cyrill-k: Backbone path properties relate to path elements not within the same domain as an endpoint considering them, thus, in the backbone from the endpoint's point of view.

draft-enghardt-panrg-path-properties.md
```diff
@@ -91,63 +92,62 @@ The notion of reliability depends on the property, it might be the confidence le
...
 
 
 # Dynamic Properties
 
-Dynamic Path Properties are expected to change on the timescale of milliseconds.
-They usually relate to the state of the path, such as the current end-to-end latency.
-Some of these properties may depend only on the first hop or on the access network, some may depend on the entire path.
-Properties related to a single layer 2 domain are abstracted from the used physical and link layer technology, similar to {{RFC8175}}.
+Dynamic path properties relate to a path segment with respect to the transmission of an individual packet or of a flow over this path segment.
+Properties related to a path element which constitutes single layer 2 domain are abstracted from the used physical and link layer technology, similar to {{RFC8175}}.
```
cyrill-k: Properties related to a path element which constitute a single layer 2 domain are abstracted from the used physical and link layer technology, similar to {{RFC8175}}.

draft-enghardt-panrg-path-properties.md
theri: Yep, fixed.

draft-enghardt-panrg-path-properties.md
theri: Fixed.

draft-enghardt-panrg-path-properties.md
theri: Fixed.

theri:
Thanks for the fixes!
I also changed the second "administrative domain" to just "domain", because I think the sentence applied to domains in general.
Now we have "administrative" only to explain that "Domain properties […] are usually in the same administrative domain as an endpoint considering them.", and when explicitly talking about an "administrative entity" in the corresponding property.

Also, I changed "set of path elements" to "path segment" because we have a definition for that now.

Is this ready to merge?

Issue #1, opened 2019-02-15
Formalize definition of a path


theri:
We need to provide (at least) one formal definition of what a path is.
Is it defined on a flow level, i.e., the hops that a set of packets traverses? Or only the hops that one individual packet traverses?

For this, we should look at the definitions in:

- IPPM
- ATM
- RFC 8175 (MANET Dynamic Link Exchange Protocol)

theri:
IPPM, in RFC 2330 (Framework for IP Performance Metrics), defines a path as a sequence of links and routers, with two hosts at the end. Each pair of (link, host) is a 'hop'.
So it is a unidirectional concept which does not depend on a packet, but a packet can take one specific path.
This is probably a good starting point for our definition...

cyrill-k:
I wrote this definition of a path at IETF 103 without looking at existing definitions in other IETF documents, but it is similar to the RFC2330 definition. This definition is at the flow level, but it might be more clear to define a path for individual packets.

A "path" is defined as the set of the devices and links between two endpoints that can be traversed by a set of packets at specific points in time. Depending on which layer the "path" is looked at, devices, links, and their properties might differ. Devices may be hidden, and a set of links may be abstracted into a single link. The lifetime of a "path" overlaps with the lifetime of the corresponding connection but is not restricted to it.

A "path" can be classified by timescale into a "measured path", based on concrete previous and current measurements, and a "potential path", which is a path with predicted characteristics, possibly including the reliability of such predictions.

theri:
Yes, I think it makes sense to first define the "path" on a per-packet level. 
Then as a second step, we can define multiple packets belonging to one flow (not just connection, but multiple UDP packets with the same source and destination IP address and port as well).
Then we can have packes of one flow taking the same path, and then we can have path properties change or stay the same for multiple packets within one flow.
So one packet can have a specific one-way latency along the path, and multiple packets of the same flow on the same path can have, e.g., a median one-way latency, or a Round Trip Time.

I'm not yet sure if we need a distinction into "measured" and "potential" path -- aren't these more, like, measured properties or predicted properties? 
I mean, the path is always there, no matter if packets are sent over it, right? But some properties only exist in relation to packets sent over a path?

cyrill-k:
What do you think about this terminology section?
Should we add it before the introduction? (I guess if we add it after, we need to remove any reference to path/property in the introduction and make a more high level introduction)
I tried to structure it accordingly (path -> flow -> property).
[...]

PR #11, opened 2019-06-30
Fix terminology

theri:
Changed definitions according to the suggestions in #1.

Questions:
- With this, a path is always between two hosts, but a subpath can be between a host and a node. I think this makes sense, as we are talking about "endpoints" (hosts) a lot. Do you agree?
- Is the part about the layering clear enough? I think a single path ultimately consists of just one sequence of path elements, just that some lower-layer path elements are hidden when you, e.g., only look at the transport layer. I'm not sure if "A path is defined on a specific layer" actually conveys this. Ideas? Opinions?

### DIFF
--- a/draft-enghardt-panrg-path-properties.md
+++ b/draft-enghardt-panrg-path-properties.md
@@ -99,10 +99,10 @@ Path element:
 : Either a node or a link.
 
 Path:
-: A sequence of directly adjacent path elements, alternating between nodes and links, starting and ending with a host. A path can be viewed as an abstraction on a specific layer, omitting lower layer path elements. For example, a router implementing IPv6 may be a path element on a path when considering the network layer. If this router does not implement transport layer functionality, it is hidden when a higher layer, such as the transport or application layer, is considered. In the case of multicast or broadcast, a single packet may be sent over multiple paths at once -- one path for each combination of sending and receiving host.
+: A sequence of adjacent path elements, alternating between nodes and links, starting and ending with a host. A path can be viewed as an abstraction on a specific layer, omitting lower layer path elements. For example, a router implementing IPv6 may be a path element on a path when considering the network layer. If this router does not implement transport layer functionality, it is hidden when a higher layer, such as the transport or application layer, is considered. In the case of multicast or broadcast, a single packet may be sent over multiple paths at once -- one path for each combination of sending and receiving host.
 
 Subpath:
-: Given a path, a subpath is a sequence of directly adjacent path elements of this path, starting and ending with a node.
+: Given a path, a subpath is a sequence of adjacent path elements of this path, starting and ending with a node.
 
 Flow:
 : One or multiple packets which are traversing the same subpath or path. For example, a flow can consist of all packets sent within a TCP session with the same five-tuple between two hosts, or it can consist of all packets sent on the same physical link.
--- a/draft-enghardt-panrg-path-properties.md
+++ b/draft-enghardt-panrg-path-properties.md
@@ -83,32 +83,43 @@ in Path-Aware Networking {{I-D.irtf-panrg-questions}}, which is a product of the
 
 # Terminology
 
-Path element:
-: A path element is a device (including the endpoints), or link used to connect two devices and transmit information on a specific layer.
-Path elements may exist on multiple layers (e.g., the endpoint corresponds to a path element on every layer), may be hidden on higher layers (e.g., a layer 2 switch in the local network), or a path element may be an aggregation of several path elements on a lower layer (e.g., the link connecting the endpoints on the transport layer being an aggregation of all network layer path elements).
+Node:
+: A device that implements a protocol on a specific layer, e.g., the physical layer, the link layer, the network layer, the transport layer, or the application layer.
 
-Path segment:
-: A path segment is an ordered set of path elements at the network layer that can be traversed by a packet.
+Host:
+: A node that handles packets that are explicitly addressed to itself.
+
+Router:
+: A node that forwards packets that are not explicitly addressed to itself.
+
+Link:
+: A medium or communication facility that connects two or more nodes with each other and enables them to exchange packets on a specific layer. A link can be physical, e.g., a WiFi network which connects an Access Point to one or multiple stations, or virtual, e.g., a virtual switch which connects two virtual machines hosted on the same physical machine.
+
+Path element:
+: Either a node or a link on a specific layer. For example, a router implementing IPv6 and Ethernet may be a path element on a network layer path, a link layer path, and a physical layer path. If this router does not implement transport layer functionality, it is hidden to any higher layer path, e.g., a transport layer path between two hosts. The link connecting these two hosts may be considered a virtual link and may abstract different physical links for each packet sent over the path.
 
 Path:
-: A path is defined as an ordered set of path elements at the network layer between two endpoints. A path can be traversed by a packet.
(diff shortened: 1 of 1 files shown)


draft-enghardt-panrg-path-properties.md
```diff
@@ -84,7 +84,7 @@ in Path-Aware Networking {{I-D.irtf-panrg-questions}}, which is a product of the
 # Terminology
 
 Node:
-: A device that implements a protocol on a specific layer, e.g., the physical layer, the link layer, the network layer, the transport layer, or the application layer.
+: A device that implements a protocol on a specific layer.
```
cyrill-k: I would refrain from defining a node as a device since in the case of network service chaining, each service would be considered a node, which means they would run on the same device.

Maybe we could replace "device" with "entity".

Another point that may be problematic is the restriction on a (single) specific layer, since some protocols might not clearly be part of a single layer (e.g., ICMP which is a layer 3 protocol but looks like a layer 4 protocol...). Should we maybe simply define a node as: "An entity that implements a function that takes packet(s) as inputs and outputs packet(s)" or is this too abstract?

draft-enghardt-panrg-path-properties.md
```diff
@@ -96,19 +96,19 @@ Link:
 : A medium or communication facility that connects two or more nodes with each other and enables them to exchange packets on a specific layer. A link can be physical, e.g., a WiFi network which connects an Access Point to one or multiple stations, or virtual, e.g., a virtual switch which connects two virtual machines hosted on the same physical machine.
 
 Path element:
-: Either a node or a link on a specific layer. For example, a router implementing IPv6 and Ethernet may be a path element on a network layer path, a link layer path, and a physical layer path. If this router does not implement transport layer functionality, it is hidden to any higher layer path, e.g., a transport layer path between two hosts. The link connecting these two hosts may be considered a virtual link and may abstract different physical links for each packet sent over the path.
+: Either a node or a link. For example, a router implementing IPv6 may be a path element on a path when considering the network layer. If this router does not implement transport layer functionality, it is hidden when a higher layer is considered, e.g., the transport layer.
```
cyrill-k: I would rephrase the second sentence:

If this router does not implement transport layer functionality, it is hidden when a higher layer, such as the transport or application layer, is considered.

draft-enghardt-panrg-path-properties.md
```diff
@@ -96,19 +96,19 @@ Link:
...
 
 Property:
-: A trait of one or multiple path elements of a path, or a trait of a flow with respect to one or multiple path elements. An example of a property of a link is the maximum data rate that can be sent over this link. An example of a property of a node is the administrative domain that the node belongs to. An example of a property of a flow with respect to one or multiple path elements is the one-way delay of a packet being sent from one node to another node over a subpath.
+: A trait of one or multiple path elements, or a trait of a flow with respect to one or multiple path elements. An example of a link property is the maximum data rate that can be sent over the link. An example of a node property is the administrative domain that the node belongs to. An example of a property of a flow with respect to a subpath is the aggregated one-way delay of the flow being sent from one node to another node over a subpath.
 A property is thus described by a tuple containing the set of path elements, the flow or an empty set if no packets are relevant for the property, the name of the property (e.g., maximum data rate), and the value of the property (e.g., 1Gbps).
```
cyrill-k: Maybe we should define a property on a sequence of path elements, since if we talk about one way latency between two nodes A and B, the order of the nodes changes the meaning of the property.

cyrill-k:
> Changed definitions according to the suggestions in #1.
> 
> Questions:
> 
>     * With this, a path is always between two hosts, but a subpath can be between a host and a node. I think this makes sense, as we are talking about "endpoints" (hosts) a lot. Do you agree?

Yes, I think it might be a good idea to more clearly separate paths and subpaths.

>     * Is the part about the layering clear enough? I think a single path ultimately consists of just one sequence of path elements, just that some lower-layer path elements are hidden when you, e.g., only look at the transport layer. I'm not sure if "A path is defined on a specific layer" actually conveys this. Ideas? Opinions?

Maybe instead of saying:
A path is defined on a specific layer.

We could say:
A path can be viewed as an abstraction on a specific layer, omitting lower layer path elements.

draft-enghardt-panrg-path-properties.md
theri: Agree that defining a node as a device is a problem. Similar to links, nodes should not be limited to physical entities.
About the layering, yes. Maybe all this layering complexity in our definitions is unnecessary anyway and we should get rid of it? I think there was a reason to introduce the layering in the first place, but I forgot.

In your proposed definition, you replace "protocol" with a "function", I think that's a good idea - not every node may implement a protocol in the traditional way (e.g., does a firewall implement a protocol if it filters based on headers of this protocol? I wouldn't say so.).
However, "input and output" sounds to me like a node necessarily transforms packets, but it may only send or only receive packets.
So how about "An entity that implements a function which sends and/or receives packets"?
Or maybe even just "An entity that sends and/or receives packets"? Because that's what makes the node different from the link, right?

draft-enghardt-panrg-path-properties.md
theri: Agreed, if we keep the layering at all -- I'm not too happy with the whole "when considering the XYZ layer", but I think the concept of "considering a path on XYZ layer" is better than defining a separate path for each layer.

draft-enghardt-panrg-path-properties.md
theri: Agreed.
I wanted to be able to define a property of, e.g., just some nodes but not the links between them, but I'm not sure if we actually need this, and if we do, I think we can work with it anyway.

draft-enghardt-panrg-path-properties.md
cyrill-k: We could say:
An entity that can receive packets from an incoming link and can send arbitrary packets on an outgoing link.

Does this imply that nodes process packets only in one direction? Or is it clear?

draft-enghardt-panrg-path-properties.md
cyrill-k: We can also define it as: "... containing the sequence or set of path elements", but this makes the concept more complex which is probably not a good idea.

Since we define it as a "sequence of path elements" and not "sequence of path elements alternating between links and nodes", we can still define a property for some nodes (they would just be ordered).

theri:
Thanks for the feedback and suggestions!

The node definition now talks about "processing" packets, but then gives a few examples. I'm still wondering if we need the "implements a function which" at all.

I like your proposal about the layering and I put everything layering-related into a single definition (the path) instead of having it spread across multiple definitions. 

Also, a property is now defined on one or a sequence of path elements - maybe we can allow a sequence of length 1 and simplify our definition accordingly? But that would confuse me if we didn't make it explicit, so I left it like it is now.

What do you think?

draft-enghardt-panrg-path-properties.md
theri: Phew, tricky. :D Yes, I think sequence is fine. If we must have the set, we can have it as a set of single properties, each of which is defined for a specific path element.

draft-enghardt-panrg-path-properties.md
theri: I went with a more open definition and just say "process" now. What do you think?

cyrill-k:
> The node definition now talks about "processing" packets, but then gives a few examples. I'm still wondering if we need the "implements a function which" at all.

I simplified it to:
An entity which processes packets, e.g., sends, receives, forwards, or modifies them.

> I like your proposal about the layering and I put everything layering-related into a single definition (the path) instead of having it spread across multiple definitions.

The definitions looks cleaner now. I removed "Paths can be traversed by packets".

> Also, a property is now defined on one or a sequence of path elements - maybe we can allow a sequence of length 1 and simplify our definition accordingly? But that would confuse me if we didn't make it explicit, so I left it like it is now.
> 
> What do you think?

I would also leave it as it is.

theri:
Thanks for the changes!
I just got feedback that "directly adjacent" can be reduced to "adjacent", which I changed in the draft.
Otherwise, I think this is good to go.

PR #12, opened 2019-07-04
Some terms fixed in other sections

theri:
Now that we've converged on the terminology, I did a first pass on the other sections, see the "adjust-terms" branch. 
Mostly, I replaced "endpoint" with "host" and "path segment" with "subpath". Sometimes I made "path element" more specific, i.e., link or node.
I think these are the minimal changes. Feel free to add more.


cyrill-k:
It looks already good. I only made a small change for the network function, since I think a network function should always be performed by a node.

theri:
You're absolutely right, thanks for the fix!
If there's no further changes, I'll merge this and submit the draft within the next few hours.

cyrill-k:
That's fine, thanks for submitting!

cyrill-k:
Sorry, I should have merged and not closed ;)

theri:
No worries ;)

Issue #14, opened 2019-08-07
Med comments on Introduction


theri:
Replying to Med's comments on the Introduction:

On "Because the current Internet provides an IP-based best-effort bit
pipe, hosts have little information about paths to other hosts.":
[Med6]: What is the causality effect here?
I think here we intend to say that, in the current Internet, the path and its services are transparent to hosts, so hosts only know something is reachable, but they do not get all the details of via which networks, they do not necessarily get details on the services (e.g., QoS).
Do you have a proposal of what text to change here?

On "A Path Aware Network exposes information about one or multiple paths
through the network":
[Med7]: I do still believe this is a weird term: networking is about manipulating paths, by essence!
Well, hosts do not usually directly manipulate paths, unless they do source routing, right? Anyway, the term "Path Aware Network" is taken from the questions draft, which is an adopted Research Group document, so I wouldn't change it and introduce even more terms in the process.
[Med8]: Difficult to parse
Will see how we can rephrase this.

Crossed out: "[It is impossible to provide an exhaustive list of path properties], as with every new technology and protocol, novel properties might become relevant"
Is the comment here that the crossed out part does this not work as an explanation of the first part of the sentence? Why not?

On "[Traffic policies: … Such policies can allow or disallow sending traffic over specific
networks or nodes":
[Med9]: This is about path selection covered in the 3rd bullet.
True, this can be a case of path selection - or you may find out that your policy does not allow you to send your traffic at all. The latter case is not covered by our current definition of path selection, so I think mentioning it here makes sense. But we should rephrase this sentence to make it more clear. Any suggestion?

On "[select a protocol depending on the capabilities of the] on-path devices":
[Med10]: I agree with this example, but I have the following comments: The selection of the protocol is not only specific to on-path devices, but it can be the other way around: It is the selection of the protocol which may infer the devices that will be involved in a path. Consider the example of the 0-RTT Convert Protocol which is involved in a path only when invoked by a host; such invocation will lead to the use of MPTCP or TCPinc while this Is not doable with the default path. Another example of traffic policies is a connection which may be composed of
multiple streams, each stream with specific requirements. A host may decide to invoke a given service function (e.g., transcoding) only for some streams while others are not processed by that service function.
True, there can be more complex interactions. Do you have a suggestion of what text to add here?

On "the monetary cost [of the link]":
[Med11]: This is linked to a network, not a path.
[...]

PR #20, opened 2019-10-22
Rephrase and restructure introduction

theri:
This addresses most comments in #14.

As a reference for architecture, we now refer to the questions draft, because that draft first what a path-aware network is.
Then, our introduction now explains the structure of the draft: Terminology, classification of properties, then some examples of properties.

For the examples of properties, we have use cases to motivate them. As these use cases became rather long and were not really suitable to be bullet points anymore, I split them into subsections.
Also, I reshuffled them, so "traffic policies" became "traffic configuration", because that's really what our examples are doing. I wanted to make the distinction between path selection and traffic configuration clear - path selection decides between different paths through the network, and then when a path has been selected, we can configure this path, which is traffic configuration. Therefore, the firewall example moved to "path selection".

Our old introduction contained a lot of text on classification of path properties. I kept this text, but moved it down after the terminology, because I think it fits better there.
Later, we'll have to think more about this classification, e.g., the domain properties issue, but I'd say this is out of scope for this PR now. ;)

What do you think?

### DIFF
--- a/draft-enghardt-panrg-path-properties.md
+++ b/draft-enghardt-panrg-path-properties.md
@@ -29,6 +29,8 @@ informative:
 
     I-D.irtf-panrg-questions:
 
+    I-D.ietf-tcpm-converters:
+
     RFC8175:
 
     RFC5693:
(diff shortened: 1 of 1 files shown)


cyrill-k:
I generally like the new structure more as it feels cleaner than before.
Maybe we could consider moving the use case section after the terminology since we use expressions such as "path" in the use case section?

Our use case section grew quite large which might be a bit of an issue since we do not want to give the impression that we provide an exhaustive list of possible use cases for path-aware networks.
But I think we should keep the text and ask during the next PANRG meeting what the audience thinks about this (it is easy to shorten these sections if necessary).

Regarding the classification, I'm in favour of removing the classification into domain properties and backbone properties and simply have dynamic/non-dynamic properties.
But I agree that we should do this in another PR.

theri:
Thanks for the review!

I moved the use cases below the terminology section because I agree that this should be a separate section and it works much nicer with our structure this way.
The use cases section now includes the disclaimer that this is not an exhaustive list because new properties may become relevant, I think that sentence fits best here.
I'll ask PANRG about the use case section at IETF 106 if we can get beyond the Terminology section. ;)

theri:
Merged this so I can base another PR on it -- Let's continue the discussion there.

PR #21, opened 2019-10-23
Change definitions based on feedback

theri:
This PR updates our terminology in line with feedback from IETF 105 and Med's comments.
See also issue #1.

A node can now be an entire AS -- can a node also be something like an amorphous collection of path elements in case of partial visibililty? (open question)

A property can now also apply to a set of subpaths -- I wanted this because the sequence of path elements on which a packet is sent can change frequently, even if the endpoints stay the same.
Do you agree?

### DIFF
--- a/draft-enghardt-panrg-path-properties.md
+++ b/draft-enghardt-panrg-path-properties.md
@@ -96,42 +96,41 @@ in Path-Aware Networking {{I-D.irtf-panrg-questions}}, which is a product of the
 # Terminology
 
 Node:
-: An entity which processes packets, e.g., sends, receives, forwards, or modifies them.
+: An entity which processes packets, e.g., sends, receives, forwards, or modifies them. A node may be physical or virtual, e.g., a machine or a service function. A node may also be the collection of multiple entities which, as a collection, processes packets, e.g., an entire Autonomous System (AS).
 
 Host:
-: A node that processes packets that are explicitly addressed to itself.
-
-Router:
-: A node that processes packets that are not explicitly addressed to itself.
+: A node that generally executes application programs on behalf of user(s), employing network and/or Internet communication services in support of this function, as defined in {{?RFC1122}}.
 
 Link:
-: A medium or communication facility that connects two or more nodes with each other and enables them to exchange packets. A link can be physical, e.g., a WiFi network which connects an Access Point to stations, or virtual, e.g., a virtual switch which connects two virtual machines hosted on the same physical machine.
+: A medium or communication facility that connects two or more nodes with each other. A link enables a node to send a packet such that another node receives the packet.
+Links can be physical, e.g., a WiFi network which connects an Access Point to stations, or virtual, e.g., a virtual switch which connects two virtual machines hosted on the same physical machine. A link is unidirectional and the collection of two links between the same nodes in different directions can be used to model bidirectional communication.
 
 Path element:
 : Either a node or a link.
 
 Path:
-: A sequence of adjacent path elements, alternating between nodes and links, starting and ending with a host. A path can be viewed as an abstraction on a specific layer, omitting lower layer path elements. For example, a router implementing IPv6 may be a path element on a path when considering the network layer. If this router does not implement transport layer functionality, it is hidden when a higher layer, such as the transport or application layer, is considered. In the case of multicast or broadcast, a single packet may be sent over multiple paths at once -- one path for each combination of sending and receiving host.
+: A sequence of adjacent path elements over which a packet can be transmitted, starting and ending with a node. Note that paths are time-dependent, i.e., the sequence of path elements over which packets are sent between two nodes may change frequently. Furthermore, an entity may have only partial visibility of the path elements that comprise a path.
 
 Subpath:
-: Given a path, a subpath is a sequence of adjacent path elements of this path, starting and ending with a node.
+: Given a path, a subpath is a sequence of adjacent path elements of this path.
 
 Flow:
-: One or multiple packets which are traversing the same subpath or path. For example, a flow can consist of all packets sent within a TCP session with the same five-tuple between two hosts, or it can consist of all packets sent on the same physical link.
+: An entity made of packets to which the traits of a path or set of subpaths may be applied in a functional sense. For example, a flow can consist of all packets sent within a TCP session with the same five-tuple between two hosts, or it can consist of all packets sent on the same physical link.
 
 Property:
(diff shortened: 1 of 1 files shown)


draft-enghardt-panrg-path-properties.md
```diff
@@ -96,42 +96,41 @@ in Path-Aware Networking {{I-D.irtf-panrg-questions}}, which is a product of the
...
 
 Path element:
 : Either a node or a link.
 
 Path:
-: A sequence of adjacent path elements, alternating between nodes and links, starting and ending with a host. A path can be viewed as an abstraction on a specific layer, omitting lower layer path elements. For example, a router implementing IPv6 may be a path element on a path when considering the network layer. If this router does not implement transport layer functionality, it is hidden when a higher layer, such as the transport or application layer, is considered. In the case of multicast or broadcast, a single packet may be sent over multiple paths at once -- one path for each combination of sending and receiving host.
+: A sequence of adjacent path elements over which a packet can be transmitted, starting and ending with a node. Note that paths are time-dependent, i.e., the sequence of path elements over which packets are sent between two nodes may change frequently. Furthermore, an entity may have only partial visibility of the path elements that comprise a path.
```
cyrill-k: I like the idea to change the path definition from "alternating path elements" to "A sequence of adjacent path elements over which a packet can be transmitted" since we address both comments on non-sensible paths (e.g., only links without nodes) and Med's comments on service function chaining.

Is it okay to remove the multicast/broadcast remark from the path definition? It was only there for clarity or do we change how multicast/broadcast is defined?

Should we add a sentence to indicate that we can look at paths using different abstraction levels (e.g., "Also note that different entities might treat path elements at different levels of abstraction."). Although I think this should be clear from the new node definition...

draft-enghardt-panrg-path-properties.md
```diff
@@ -96,42 +96,41 @@ in Path-Aware Networking {{I-D.irtf-panrg-questions}}, which is a product of the
...
 
 Subpath:
-: Given a path, a subpath is a sequence of adjacent path elements of this path, starting and ending with a node.
+: Given a path, a subpath is a sequence of adjacent path elements of this path.
```
cyrill-k: Do we still need subpath if we define path as being between two nodes (not hosts as before)?

draft-enghardt-panrg-path-properties.md
```diff
@@ -96,42 +96,41 @@ in Path-Aware Networking {{I-D.irtf-panrg-questions}}, which is a product of the
...
 
 Property:
-: A trait of one or a sequence of path elements, or a trait of a flow with respect to one or a sequence of path elements. An example of a link property is the maximum data rate that can be sent over the link. An example of a node property is the administrative domain that the node belongs to. An example of a property of a flow with respect to a subpath is the aggregated one-way delay of the flow being sent from one node to another node over a subpath.
-A property is thus described by a tuple containing the sequence of path elements, the flow or an empty set if no packets are relevant for the property, the name of the property (e.g., maximum data rate), and the value of the property (e.g., 1Gbps).
+: A trait of one or a sequence of path elements, or a trait of a flow with respect to one or multiple subpaths. An example of a link property is the maximum data rate that can be sent over the link. An example of a node property is the administrative domain that the node belongs to. An example of a property of a flow with respect to a subpath is the aggregated one-way delay of the flow being sent from one node to another node over this subpath.
```
cyrill-k: Did you change the Property definition (on one or multiple subpaths) to allow properties on links between multiple nodes (e.g., wireless links?) Or is there some other motivation?

cyrill-k: Sorry I just now saw your comment on this pull request...
I'm thinking that for properties between endpoints, we might need an abstract link analogous to the abstract node. Then we can model the (sub-)paths between endpoints as two nodes/hosts connected by the abstract link (e.g., 5-tuple).
This approach which associates a property with multiple (sub-)paths could work as well but might have the issue that it is not clear when the property held for which of the (sub-)paths.

cyrill-k:
> A node can now be an entire AS -- can a node also be something like an amorphous collection of path elements in case of partial visibililty? (open question)

I think defining an abstract node which can be a collection of entities is useful but then the question comes up if we even need links at all or if links are just abstract nodes that only forward packets?

draft-enghardt-panrg-path-properties.md
theri: About multicast/broadcast: It was only there for clarity, but I think this should be explicitly stated somewhere. I wasn't sure it had to be in the path definition, so I took it out there and added it to my list of "things to check before submitting", so I was planning to add it somewhere else. However, on second thought, maybe the path definition is the best place to add it after all... I think I'll just re-add it to the definition for now. Maybe we'll find a better place for this kinds of clarifications later on - maybe we'll even add a figure? But that's future work.

About the different abstraction levels, similarly, I think this should be somewhere. I'm not quite sure yet where, but I think this is related to partial visibility. The path definition already says "an entity may have only partial visibility of the path elements that comprise a path", so I'll just add something on the abstraction level there for now.

draft-enghardt-panrg-path-properties.md
theri: Right, I see. You're right, having multiple subpaths in there might get too confusing, so I'll change it back. A property should apply to one or a sequence of path elements (and maybe a flow), so only one subpath.
I like your idea of an abstract link: For cases in which this subpath changes frequently, either we care and have insight into these path elements, or we do not care and just use an abstract link.

draft-enghardt-panrg-path-properties.md
cyrill-k: I think the updated path definition in 15964a38fd16cadc06a88f04add65487e0499420 looks good now. I'm a bit worried that the abstraction levels are not clearly defined but I also couldn't come up with a better way to describe what we mean...

cyrill-k:
> About viewing a link as just another node that forwards packets, this sounds kind of confusing to me and I don't think I agree:
> I think "forward packets" means different things for nodes and for links - nodes perform routing and switching, links connect one node to another. That's not the same thing, but we don't use "forward" for links in our definition, so we should be fine.

I agree with you. It was just going through my head when I saw the definition of an abstract node. :)

draft-enghardt-panrg-path-properties.md
theri: Agree, let's keep it like this for now and see what the Research Group thinks.

theri:
Thanks for taking another look!
So it looks like we can merge this for now?

Issue #16, opened 2019-08-07
Med comments on Backbone properties


theri:
On "Typically, backbone properties are less accessible to a host than domain properties, due to the potential increased distance and the lack of pre-existing trust or contractual relationship."
[Med39]: Not sure about this.
Not sure if "typically" is the right word here either, but, similar to the part with [Med33], the intention is to say that this might be the case.

On "hosts are less likely to be able to influence which path elements form their path in the backbone, as well as their properties."
[Med40]: This may contradict with slicing use cases.
Sounds interesting. Could you point me to a specific reference, please?

On "Presence of a certain [CROSSED: network] service function on the path":
[Med41]: To be aligned with RFC7665
Agreed.

On "Administrative Entity: The administrative entity, e.g., the AS, to which a path element or subpath belongs."
[Med42]: This is an administrative domain, not * entity *
Isn't an AS also an entity? But yes, maybe we could have the administrative domain be a property instead?

On "Disjointness":
[Med43]: you may also refer to the PCE RFCs which defines also the notion of near disjointness
I found RFC 4655 saying "The path computation request may include a significant set of requirements, including the following: […] the number of disjoint paths required and whether near-disjoint paths are acceptable", but it does not define what near-disjoint is. Is there a specific degree of disjointness that is generally still considered "near"? Anyways, I think if a host or other entity knows about the degree of disjointness, it can decide whether this is "disjoint" enough. Maybe we want to add a reference to PCE as an example use case.

theri:
Replies to the e-mails:

To [Med39]:
"[Med] OK, but still don't see the need to have such assertion included. "

To [Med42]:
"[Med] The administrative entity is the network operator of that AS. 

[TE] Right. We may need a different word here."

Then my suggestion:
"[TE] But yes, maybe we could have the administrative domain be a property instead?

[Med] That would make sense. For example, an application can decide to select a path that does not cross a given AS or better, an application can indicate a hint to an upstream network to avoid forwarding data via paths including a given AS.

[TE] Yes, that's the primary use case we had in mind, so we'll make sure it's supported."

theri:
Replies to the e-mails:

To [Med39]:
"[Med] OK, but still don't see the need to have such assertion included. "

To [Med42]:
"[Med] The administrative entity is the network operator of that AS. 

[TE] Right. We may need a different word here."

Then my suggestion:
"[TE] But yes, maybe we could have the administrative domain be a property instead?

[...]

Issue #15, opened 2019-08-07
Med comments on Domain properties


theri:
On "3. Domain properties":
[Med28]: I’m afraid this section (and section 4) implies architectural considerations which may not be appropriate in a terminology document.
As the scope of this document is to answer the first question of the Questions draft (Section 2.1 in draft-irtf-panrg-questions), "how are path properties defined and represented?", yes, this is a terminology document, but I wonder if possible classification of path properties is in scope. Here, our intention was to introduce such classification (e.g., dynamic and non-dynamic) to make it easier to reason about different properties, their usefulness, etc. However, I'm not sure if the classification into "domain" and "backbone" properties is indeed the right way to go. In the draft we give a number of reasons why it may matter if a path element is in the same administrative domain as a host or not. I'm sure there's a number of considerations here that may already touch upon other questions of the Questions draft, e.g., the second question of Discoverability, Distribution, and Trustworthiness of path property. However, the second question already asks how an endpoint can get access to trustworthy path properties, which we do not attempt to answer in this draft. So I'm not sure what the right trade-off here is between focusing the document on terminology and definitions vs. discussing considerations of where path elements may be located and who may control them. I'm very interested to get guidance from the Research Group here.

On "the first few hops":
[Med29]: 2, 10 ?
We kept this intentionally vague to cover both cases in we really consider only the first hop and cases in which we consider the entire access network behind that hop as well. I think deciding whether we want a clear definition here depends resolving the issue raised by the comment [Med28] above.

On "the same administrative domain as a host considering them."
[Med30]: L2, L3, ?
Personally I mostly think about domains in this document as everything covered by the same AS, but yes, it would be very interesting to come up with a clear definition of domain here.
For example, PCE defines a domain as a "collection of network elements within a common sphere of address management or path computation responsibility, can be IGP areas, ASes, multiple ASes within a Service provider network. may also exist as sub-domains of areas or ASes.". PANRG is more broad, but maybe we can come up with something similar here?
Again, I think deciding whether we want a clear definition here depends resolving the issue raised by the comment [Med28] above.

[Med31]: This would exclude then a host connected to a UE connected to a PLMN or a CPE connected to an enterprise or provider networks. These elements do not belong to the same
administrative entity.
[...]

PR #22, opened 2019-10-30
Remove domain/backbone classification, address #15 and #16

theri:
This removes the domain/backbone classification, leaving only dynamic and non-dynamic properties.

Right now, dynamic properties relate to the transmission of a specific flow and non-dynamic are everything else - not sure yet if the naming still makes sense this way.
Other ideas for definitions or names?

Anyway, I think that this addresses the remaining comments in #15 and #16.

### DIFF
--- a/draft-enghardt-panrg-path-properties.md
+++ b/draft-enghardt-panrg-path-properties.md
@@ -164,7 +164,7 @@ The merit of a momentary measurement of a dynamic path property diminishes great
 
 Non-dynamic path properties relate to individual path elements or subpaths.
 From the point of view of a host, such path properties may relate to path elements close to the host, i.e., within the first few hops, or they may include path elements far from the host, e.g. list of ASes traversed.
-The visibility of path properties to a specific entity may depend based on the physical or network distance, the existence of any trust or contractual relationships between the entity and the path element(s), or other factors.
+The visibility of path properties to a specific entity may depend on factors such as the physical or network distance or the existence of trust or contractual relationships between the entity and the path element(s).
 
 Furthermore, entities may or may not be able to influence the path elements on their path and their path properties.
 For example, a user might select between multiple potential adjacent links by selecting between multiple available WiFi Access Points. Or when connected to an Access Point, the user may move closer to enable their device to use a different access technology, potentially increasing the data rate available to the device.
--- a/draft-enghardt-panrg-path-properties.md
+++ b/draft-enghardt-panrg-path-properties.md
@@ -153,17 +153,11 @@ For example, a 0-RTT Transport Converter {{I-D.ietf-tcpm-converters}} will be in
 
 This Section gives some examples of Path Properties which may be useful, e.g., for the use cases described in {{use-cases}}.
 
-Path properties may be relatively dynamic or non-dynamic.
-Dynamic path properties relate to the transmission of an individual packet or of a flow over a subpath, e.g., the one-way delay of a packet sent over a specific path.
-Non-dynamic path properties relate to individual path elements or subpaths.
-Usefulness over time is fundamentally different for dynamic and non-dynamic properties.
+Path properties may be relatively dynamic, e.g., the one-way delay of a packet sent over a specific path, or non-dynamic, i.e., only change infrequently.
+Usefulness over time differs depending on how dynamic a property is:
 The merit of a momentary measurement of a dynamic path property diminishes greatly as time goes on, e.g. the merit of an RTT measurement from a few seconds ago is quite small, while a non-dynamic path property might stay relevant for a longer period of time, e.g. a NAT typically stays on a specific path during the lifetime of a connection involving packets sent over this path.
 
-
-## Non-dynamic Path Properties
-
-Non-dynamic path properties relate to individual path elements or subpaths.
-From the point of view of a host, such path properties may relate to path elements close to the host, i.e., within the first few hops, or they may include path elements far from the host, e.g. list of ASes traversed.
+From the point of view of a host, path properties may relate to path elements close to the host, i.e., within the first few hops, or they may include path elements far from the host, e.g. list of ASes traversed.
 The visibility of path properties to a specific entity may depend on factors such as the physical or network distance or the existence of trust or contractual relationships between the entity and the path element(s).
 
 Furthermore, entities may or may not be able to influence the path elements on their path and their path properties.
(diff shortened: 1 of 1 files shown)


draft-enghardt-panrg-path-properties.md
```diff
@@ -147,57 +149,39 @@ Conversely, the selection of a protocol may influence the devices that will be i
...
-# Path Property Classification
+# Examples of Path Properties
+
+This Section gives some examples of Path Properties which may be useful, e.g., for the use cases described in {{use-cases}}.
 
-Path properties may be relatively dynamic, e.g. median one-way delay of packets sent over a specific path within the last second.
-Path properties may also be non-dynamic, i.e., change less frequently, i.e., on a timescale of seconds or more.
+Path properties may be relatively dynamic or non-dynamic.
+Dynamic path properties relate to the transmission of an individual packet or of a flow over a subpath, e.g., the one-way delay of a packet sent over a specific path.
+Non-dynamic path properties relate to individual path elements or subpaths.
```
cyrill-k: I can't come up with another good naming than dynamic vs non-dynamic. I think if we use "flow", e.g., "Flow Properties", then this would indicate that other properties are not defined for a flow, which is not necessarily always the case. As an example, we defined "Monetary cost" and "Service function" as non-dynamic properties which relate to a flow.

Instead of dynamic, we could maybe label them as short-term and long-term properties and refer to the expected time for which the property is useful.

draft-enghardt-panrg-path-properties.md
```diff
@@ -212,14 +196,14 @@ Protocol Features available:
 : Whether a specific protocol feature is available over a path or subpath, e.g., Explicit Congestion Notification (ECN), or TCP Fast Open.
 
 
-# Dynamic Properties
+## Dynamic Path Properties
 
 Dynamic path properties relate to the transmission of an individual packet or of a flow over a subpath.
```
cyrill-k: As said in the first comment, maybe we should change this since this also includes some non-dynamic path properties.

draft-enghardt-panrg-path-properties.md
```diff
@@ -147,57 +149,39 @@ Conversely, the selection of a protocol may influence the devices that will be i
...
-Domain properties expose a high amount of information to hosts and strongly influence the connection behavior while there is little influence and information about backbone properties.
 
-Dynamic properties are not separated into domain and backbone properties, since most of these properties are defined for a complete path and it is difficult and seldom useful to define them on part of the path.
-There are exceptions such as dynamic wireless access properties, but these do not justify separation into different categories.
+## Non-dynamic Path Properties
 
+Non-dynamic path properties relate to individual path elements or subpaths.
+From the point of view of a host, such path properties may relate to path elements close to the host, i.e., within the first few hops, or they may include path elements far from the host, e.g. list of ASes traversed.
+The visibility of path properties to a specific entity may depend based on the physical or network distance, the existence of any trust or contractual relationships between the entity and the path element(s), or other factors.
```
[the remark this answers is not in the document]
theri: Sounds much better! Changing it.

draft-enghardt-panrg-path-properties.md
theri: You're right, both dynamic and non-dynamic properties may relate to the transmission of a specific flow.
What I meant here was: Dynamic properties express something related to the performance of transmitting this flow over these path elements. Right? All of the current dynamic properties directly express some kind of performance measurement, like how long the transmission took or how much data can be transmitted over time. And none of the non-dynamic properties directly expresses some performance measurement.
So maybe something like "performance properties" or "transmission performance properties"?

I think we used "dynamic" or "short-term" for these properties but actually meant that a specific measurement of these properties is a momentary thing, and performance changes over time, so the next measurement might be different.
However, I'm rethinking "dynamic" or "short-term" because the Link Capacity is actually not something I'd expect to change all the time. For a single link, it might not actually change that much (like for an Ethernet cable), and for a path, it might only change when the path changes, similar to what we have for "non-dynamic" properties right now.

Of course, our last resort could be to just get rid of all the classifications for now, if we can't decide ;)

draft-enghardt-panrg-path-properties.md
```diff
@@ -147,57 +149,39 @@ Conversely, the selection of a protocol may influence the devices that will be i
...
-\[TODO: Remove split between domain and backbone properties? Or at least remove some of those claims? \]
-
-Non-dynamic properties are further separated into (local) domain properties related to the first few hops of the connection, and backbone properties related to the remaining hops.
-Domain properties expose a high amount of information to hosts and strongly influence the connection behavior while there is little influence and information about backbone properties.
 
-Dynamic properties are not separated into domain and backbone properties, since most of these properties are defined for a complete path and it is difficult and seldom useful to define them on part of the path.
-There are exceptions such as dynamic wireless access properties, but these do not justify separation into different categories.
+## Non-dynamic Path Properties
 
+Non-dynamic path properties relate to individual path elements or subpaths.
```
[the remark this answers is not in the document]
theri: True... but how to phrase this then?

cyrill-k:
Maybe getting rid of all the classifications is the easiest thing to do, because every time I try to come up with another classification, there is at least one property that applies to both classifications...
For example if we simply say relevant for performance vs not relevant for performance, then we have the issue that (almost) all properties are potentially relevant for the performance.

But I think classifying based on how a property **expresses** the transmission performance might be a good idea. Since for a transmission, the properties (loss, delay, link usage) are what's mostly important and not the underlying properties such as link capacity, disjointness, or access technology that lead to the experienced performance. So we could maybe have _transmission performance properties_ (loss, delay, link usage) vs _other properties_. Then we could put the _other properties_ first and after that add a subsection for the _transmission performance properties_.
What do you think about this?

theri:
Yes, I think saying that the property expresses transmission performance, so it's a transmission performance property, sounds like a possible classification.
Having the other category be "other" properties sounds kind of awkward though, especially having an "other" category that comes first. Or do you mean we don't make the "other" category explicit and simply say that these are examples of path properties, and then there's path properties that specifically express transmission performance? That could work, but then I wouldn't put them in a subsection, because having only a single subsection doesn't really make sense.
So, should we get rid of the subsections, have text about the "other" properties first, then list them, then have some text about the transmission performance properties, and list them?

cyrill-k:
Yes that is what I had in mind to at least provide some structure in this section instead of simply dumping a list of properties ;)
I agree that a single subsection looks wierd so separating the performance properties by some text is a good idea.

theri:
Okay, all properties are now in the same section, without any subsections, but separated by text.
I chose to keep some text on how properties may be more or less dynamic, with the idea that it's not necessary to have a clear distinction, in the same way properties may apply to path elements close or less close to a host, but we don't have to define "close".
Do you think this works?

Also, I removed one sentence from the Introduction because we don't actually have a separate section talking about path property classification anymore.

theri:
Oh, by the way, I got rid of the sentence "Some dynamic properties are defined in different directions for the same path element, e.g., for transmitting and receiving packets." because links are unidirectional now and a single packet is only sent in one direction. Do you agree?

Also, I'm wondering if we want to get rid of the very last property, "Congestion" - I'm not sure if we really need this in our list and I'm not sure it belongs with the transmission performance properties.

cyrill-k:
The document looks good now. I removed the congestion path property since ECN is already mentioned before and added ethernet MTU as a non-dynamic property.

I agree that it is unnecessary to explicitly mention the direction of a property.

I was thinking whether we want to move "Link Capacity" away from the transmission performance path properties since we are usually interested in the "Link Usage" for the performance.
But since these two properties fit together nicely, I would leave it as it is now.

cyrill-k:
Btw., did the PANRG chairs reply whether we should upload the draft as draft-irtf... or draft-enghardt...?

theri:
Thanks for the further changes, I agree this is ready to go now.
I talked to the chairs and we'll submit it as draft-enghardt again this time, then confirm the adoption in the PANRG session.
I'll get the submission ready this afternoon.

Issue #23, opened 2019-12-05
Include editorial changes and some new definitions based on Med's review of -03


theri:
Include some of the changes in https://github.com/boucadair/IETF-Drafts-Reviews/blob/master/draft-enghardt-panrg-path-properties-03-rev%20Med.pdf

In the following places we may change the suggested inputs from the review:

In the "Node" definition, we would like to highlight that a node can be physical or virtual, so we decided we wanted to give examples for both - so we'll put in "e.g., a physical machine or a service function" (and not virtual machine), because the service function is an example for a virtual node, so we'll have the machine be a physical one.

Towards the end of the "Path" definition, there's a sentence on path visibility. We would like to make sure that the visibility aspect is clear in our definitions, but we are not sure yet whether adding this particular sentence is the best way to do it, as we already have a sentence on visibility right before. We'll consider how to phrase and possibly consolidate our definition in the next revision. 

(Not include the discovery part)

The definitions of Reverse path and Symmetric path seem useful to us as part of a base terminology for path properties, so we'll include them.

Thank you for the input on the "Use cases" section. We are going to revise this section based on input we got at IETF 106, and we will consider your feedback as well when doing so. As we revise our use cases section, we will consider whether to put any reference to them into the abstract or not.

For the definition of the "Transport Protocols available" property, again, we are not sure we want to go into detail about how to discover or cache this property, so we would opt for less detail here.

PR #30, opened 2020-01-29
Editorial and definition changes based on Med's review of -03

theri:
This should cover everything discussed in #23 except the use cases section, which we tackle in another issue.
I added the definitions we promised to add while limiting our scope to definitions, i.e., not trying to go into how paths and path properties can be discovered. That's also why I cut some details for the "Transport Protocols available" property: Sure, a host can cache this because it's visible locally, but also based on probing as Med correctly pointed out, but let's try to avoid going into these different possibilities here.

### DIFF
--- a/draft-enghardt-panrg-path-properties.md
+++ b/draft-enghardt-panrg-path-properties.md
@@ -180,7 +180,10 @@ Administrative Domain:
 : The administrative domain, e.g., the IGP area, AS, or Service provider network to which a path element belongs.
 
 Disjointness:
-: For a set of two paths or subpaths, the number of shared path elements can be a measure of intersection (e.g., Jaccard coefficient, which is the number of shared elements divided by the total number of elements). Conversely, the number of non-shared path elements can be a measure of disjointness (e.g., 1 - Jaccard coefficient). We typically say that a bidirectional path is symmetric if the path and reverse path have a Jaccard coefficient of one. A multipath protocol might use disjointness as a metric to reduce the number of single points of failure.
+: For a set of two paths or subpaths, the number of shared path elements can be a measure of intersection (e.g., Jaccard coefficient, which is the number of shared elements divided by the total number of elements). Conversely, the number of non-shared path elements can be a measure of disjointness (e.g., 1 - Jaccard coefficient). A multipath protocol might use disjointness as a metric to reduce the number of single points of failure.
+
+Symmetric Path:
+: Two paths are symmetric if a path and its reverse path consist of the same path elements, but in reverse order, i.e., the paths have a Jaccard coefficient of one.
 
 Path MTU:
 : The maximum size, in octets, of an IP packet that can be transmitted without fragmentation.
--- a/draft-enghardt-panrg-path-properties.md
+++ b/draft-enghardt-panrg-path-properties.md
@@ -56,7 +56,8 @@ informative:
 Path properties express information about paths across a network and the services provided via such paths.
 In a path-aware network, path properties may be fully or partially available to entities such as hosts.
 This document defines and categorizes path properties.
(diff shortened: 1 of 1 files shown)


draft-enghardt-panrg-path-properties.md
```diff
@@ -73,20 +74,26 @@ Finally, the document lists several path properties that may be useful for the m
...
+: A sequence of adjacent path elements over which a packet can be transmitted, starting and ending with a node. A path is unidirectional. Paths are time-dependent, i.e., the sequence of path elements over which packets are sent from one node to another may change. A path is defined between two nodes. For multicast or broadcast, a packet may be sent by one node and received by multiple nodes. In this case, the packet is sent over multiple paths at once, one path for each combination of sending and receiving node; these paths do not have to be disjoint. Note that an entity may have only partial visibility of the path elements that comprise a path and visibility may change over time. Different entities may have different visibility of a path and/or treat path elements at different levels of abstraction.
+
+Reverse Path:
+: The path that is used by a remote node in the context of bidirectional communication.
+
+Symmetric Path:
+: Denotes the case where the reverse path can be inferred from the path. For example, if the path is composed of path elements (PE1, PE2, PE3), if the path is symmetric, the reverse path is (PE3, PE2, PE1).
```
cyrill-k: I like the first sentence but I'm not sure if reverse([PE1,PE2,PE3]) = [PE3,PE2,PE1] is always true since a path element might not have a reverse path element, especially for virtual nodes (e.g., node that drops packets in one direction => no corresponding reverse node).

Maybe we could say that there are different degrees of symmetry, e.g., only a subset of nodes is symmetric (I could imagine an IP symmetric path where nodes describing IP routers that decrease the TTL are symmetric).

draft-enghardt-panrg-path-properties.md
```diff
@@ -73,20 +74,26 @@ Finally, the document lists several path properties that may be useful for the m
...
+: A sequence of adjacent path elements over which a packet can be transmitted, starting and ending with a node. A path is unidirectional. Paths are time-dependent, i.e., the sequence of path elements over which packets are sent from one node to another may change. A path is defined between two nodes. For multicast or broadcast, a packet may be sent by one node and received by multiple nodes. In this case, the packet is sent over multiple paths at once, one path for each combination of sending and receiving node; these paths do not have to be disjoint. Note that an entity may have only partial visibility of the path elements that comprise a path and visibility may change over time. Different entities may have different visibility of a path and/or treat path elements at different levels of abstraction.
+
+Reverse Path:
+: The path that is used by a remote node in the context of bidirectional communication.
```
cyrill-k: I like this definition as it is generally applicable.

draft-enghardt-panrg-path-properties.md
theri: I would have thought that a path is symmetric if the return path consists of the same path elements, but just in the reverse order/direction, but yes, there might be cases where this is not entirely true. Also, the "same link in the reverse direction" might be a problem sometimes.

Maybe here we could use disjointness, which we define later. So maybe we could move "Symmetric path" to the examples section because it's essentially a special case of disjointness, right?

cyrill-k: I like this suggestion and added it in 23dd972.
What do you think about the changes?

theri: Thanks for moving the definition! I think it makes more sense as a separate item, so I separated it out from the Disjointness definition.
Also, I merged the 'master' branch into this one to resolve the merge conflicts in our Path definition and Access Technology definition.

cyrill-k:
I think this looks fine now, so from my side, we can merge it

PR #31, opened 2020-01-29
Refactor use cases, fix #29

### LINKED ISSUE #29: Use cases
At IETF 106, Mirja had some comments on our use cases:

- Performance monitoring: She reads this as "how is my network performing right now", but she doesn't think that's what we are looking for in PANRG. We are more looking at general performance properties, like over the last week. She thinks this group is scoped on more long-term, not last second. I replied that we can have both short-term and long-term, but somehow for her we are talking about how we measure, and we should be talking more about definitions instead (if I understood her correctly). So I guess we should rethink this use case with Mirja's comment in mind.

- Traffic configuration: Talks a lot about protocol selection. Mirja suggested that we should separate out protocol selection as its own use case, which is "simple" and which we understand, and then maybe extend our use cases later.

theri:
Based on the comments we got at IETF 106 and on Med's review, I refactored the use cases.
I removed Performance Monitoring and Enhancement because monitoring was mostly talking about how to get path properties (which is out of scope for our draft) and then enhancement can take different forms: Path selection, protocol selection/configuration, service invocation, or perhaps others.
Here I'm just wondering whether to include a reference to LOOPS somewhere, e.g., a path element adding some FEC on some encapsulated traffic based on path properties. But we don't really have an adopted draft for that yet, while other examples such as TCP converters and performance routing are already adopted IETF documents.

I reshuffled the Path Selection section a bit because I think the part of when you can do path selection  should come earlier, and also the warning to be aware of their limitations, before going into examples.

Finally I split Protocol Selection and Service Invocation (named based on a suggestion by Med), because I think these are really separate things to do to your traffic and/or path.

Please let me know what you think. :)

BTW, I'd like to get a new revision out soon-ish (within the next few weeks), i.e., before my planned relocation.
If you want to do a PR for any of the remaining issues yourself before then, feel free to self-assign an issue.

### DIFF
--- a/draft-enghardt-panrg-path-properties.md
+++ b/draft-enghardt-panrg-path-properties.md
@@ -31,23 +31,29 @@ informative:
 
     I-D.ietf-tcpm-converters:
 
-    RFC8175:
+    I-D.ietf-quic-transport:
 
-    RFC5693:
+    I-D.ietf-idr-performance-routing:
 
-    RFC5136:
-
-    RFC7679:
+    RFC3357:
 
     RFC3393:
 
-    RFC7680:
+    RFC4271:
 
-    RFC3357:
+    RFC5136:
+
+    RFC5693:
 
     RFC6534:
 
     RFC7665:
+
+    RFC7679:
+
+    RFC7680:
+
+    RFC8175:
 
     ANRW18-Metrics: DOI.10.1145/3232755.3232764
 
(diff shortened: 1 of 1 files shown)


cyrill-k:
I made a commit a6d104d with some small textual changes and I added the following to the path selection use case:

Path selection based on measurements might not just be useless but could also be harmful for the stability.

theri:
Thanks for the commit, I like your changes!
So, can we merge this already?

PR #32, opened 2020-02-12
Clarifications, fix #24

### LINKED ISSUE #24: Clarifications based on review by Luis M. Contreras
On "access technology":

"As a path consists of a sequence of path elements, which can be nodes or
links, our definitions do not enforce specifying an interface or
sub-interface as part of the path.

We would like to make this clearer in the draft and we are wondering if the question comes from the choice of the term "access technology". We think what we actually mean here just the "link type" on the physical or link layer, no matter if in the access network or anywhere else on the path. Does renaming the property help? Or do you think the definition of the property has to be changed?

Otherwise, maybe it would help to say explicitly that defining a property, such as "link type", does not mean that we require each entity to specify or know about this property. So, if an entity does not know about the link type or the exact interfaces involved in a path, that is not a problem, the entity can still express the path based on all path elements it knows or cares about.

Does this help?"

On being flexible to include different levels of abstractions:

"Our intention is for the definitions to be flexible enough to accomodate both [a path can be simply equivalent to a route (i.e., a list of links) or something much more complex, including e.g. slots in FlexEth connections or ODUs in OTN, aggregated links in LAG configurations, etc.]

For example, using our current definitions it should be possible to express a complex path including slots in FlexEth connections, or aggregating multiple links (physical or virtual) to one virtual link.

We will consider how to make this clearer in the next revision."

theri:
Explicitly stated some points that people keep misunderstanding in response to #24:

- Make scope of the document explicit
- Add examples to our path definition to make the concept of "different levels of abstractions" clearer
- Refactored Access Technology to make it clearer how this relates to either links or nodes


draft-enghardt-panrg-path-properties.md
```diff
@@ -86,7 +88,7 @@ Path element:
...
+: A sequence of adjacent path elements over which a packet can be transmitted, starting and ending with a node. Paths are time-dependent, i.e., the sequence of path elements over which packets are sent from one node to another may change frequently. A path is defined between two nodes. For multicast or broadcast, a packet may be sent by one node and received by multiple nodes. In this case, the packet is sent over multiple paths at once, one path for each combination of sending and receiving node. Note that an entity may have only partial visibility of the path elements that comprise a path, and entities may treat path elements at different levels of abstraction. For example, a path may be given as a sequence of physical nodes and the links connecting these nodes, or it may be given as a sequence of logical nodes such as a sequence of ASes or an Explicit Route Object (ERO). Similarly, the representation of a path and its properties, as it is known to a specific entity, may be complex and include details on the used physical layer technology, or it may be abstract and only include a route to a specific destination.
```
cyrill-k: I would change the last sentence to "Similarly, the representation of a path and its properties, as it is known to a specific entity, may be more complex and include details about the physical layer technology, or it may be more abstract and only consist of a specific source and destination."

Since saying "only include a route" does not necessarily mean that the route is abstract and also we didn't define what a route is.

theri: Agree, "only include a route" is not specific enough here.
What I meant here is the information that a specific destination is reachable. I think your proposed sentence works, because if a path consists of a specific source and destination, this implies that it is reachable, otherwise it wouldn't be a path according to our definition.
However, if we want to be really really clear, how about "only consist of a specific source and a destination which is known to be reachable from that source"?

cyrill-k: Ok, I changed the sentence

PR #33, opened 2020-02-19
Add transparency property

cyrill-k:
I defined transparency as a property of a service function.

I'm not sure if this restricts the usefulness of the transparency property.
Are there cases where transparency is relevant for nodes/links that are not service functions?

Do you have other suggestions how to define it?


cyrill-k:
I like the change from service function to node since it is more applicable.
But I think saying if a node records meta-info about a protocol it cannot be transparent to that protocol is perfect since the node might simply parse the complete packet but not act on the meta-info. Maybe we can say if the node doesn't act based on the meta-info recorded for a protocol, the node is transparent to this protocol.

I revised the text but I'm not too happy with the sentence yet. If you see a way to describe it more nicely, feel free to change it.

PR #38, opened 2020-08-30
Add examples to host definition

cyrill-k:
Add examples (web browser for client node and web server for server node) to host definition

### DIFF
--- a/draft-irtf-panrg-path-properties.md
+++ b/draft-irtf-panrg-path-properties.md
@@ -86,7 +86,7 @@ Node:
 
 Host:
 : A node that generally executes application programs on behalf of user(s), employing network and/or Internet communication services in support of this function, as defined in {{?RFC1122}}.
-Note that hosts include both client nodes (e.g., running a web browser) and server nodes (e.g., deploying a web server).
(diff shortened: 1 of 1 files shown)


draft-irtf-panrg-path-properties.md
```diff
@@ -86,6 +86,7 @@ Node:
 
 Host:
 : A node that generally executes application programs on behalf of user(s), employing network and/or Internet communication services in support of this function, as defined in {{?RFC1122}}.
+Note that hosts include both client nodes (e.g., running a web browser) and server nodes (e.g., deploying a web server).
```
theri, line 89:
```suggestion
Note that hosts include both client nodes (e.g., running a web browser) and server nodes (e.g., running a web server).
```
"deploying" sounds confusing to me, as if the server was installing the software to another machine. For me, "deploying" is definitely not a synonym of running.

theri: If you really want "deploy", how about "[…] and server nodes (e.g., on which a web server is deployed)"?

cyrill-k: I agree that it is more clear to use running consistently

PR #39, opened 2020-08-30
Note that we do not define trust properties because we do not define …

cyrill-k:
Note that we do not define trust properties because we do not define a threat model

### DIFF
--- a/draft-irtf-panrg-path-properties.md
+++ b/draft-irtf-panrg-path-properties.md
@@ -241,8 +241,8 @@ If nodes are basing policy or path selection decisions on path properties, they 
 In order to be able to trust such path properties, nodes may need to establish a trust relationship or be able to verify the authenticity, integrity, and correctness of path properties received from another node.
 
 Security related properties such as confidentiality and integrity protection of payloads are difficult to characterize since they are only meaningful with respect to a threat model which depends on the use case, application, environment, and other factors.
-Likewise, properties for trust relations between nodes cannot be meaningfully defined without a concrete threat model.
-Such properties are orthogonal to the path terminology and path properties defined in this document, as they are tied to the communicating entities and protocols used (e.g., client and server using HTTPS, or client and remote network node using VPN) and the path is typically oblivious to these properties.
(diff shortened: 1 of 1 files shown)


draft-irtf-panrg-path-properties.md
```diff
@@ -241,6 +241,7 @@ If nodes are basing policy or path selection decisions on path properties, they
 In order to be able to trust such path properties, nodes may need to establish a trust relationship or be able to verify the authenticity, integrity, and correctness of path properties received from another node.
 
 Security related properties such as confidentiality and integrity protection of payloads are difficult to characterize since they are only meaningful with respect to a threat model which depends on the use case, application, environment, and other factors.
+Likewise, properties for trust relations between nodes cannot be meaningfully defined without a concrete threat model.
```
theri, line 244:
```suggestion
Likewise, properties for trust relations between nodes cannot be meaningfully defined without a concrete threat model, and defining a threat model is out of scope for this draft.
```

draft-irtf-panrg-path-properties.md
```diff
@@ -241,6 +241,7 @@ If nodes are basing policy or path selection decisions on path properties, they
...
 Such properties are orthogonal to the path terminology and path properties defined in this document, as they are tied to the communicating entities and protocols used (e.g., client and server using HTTPS, or client and remote network node using VPN) and the path is typically oblivious to these properties.
```
theri, line 245:
```suggestion
Properties related to confidentiality, integrity, and trust are orthogonal to the path terminology and path properties defined in this document, as they are tied to the communicating entities and protocols used (e.g., client and server using HTTPS, or client and remote network node using VPN) and the path is typically oblivious to these properties.
```
Trying to make it go well with the sentence before.
Also, this sentence is rather long, maybe we should split it? But I can't immediately see a good way to do so. Doesn't have to be in this PR.

cyrill-k: I like the changes.
i will split the long sentence in another commit

cyrill-k (commented):
Security Considerations Textual changes

cyrill-k:
I rephrased the long sentence.
What do you think?

theri (approved):
Looks good to me! Thank you.

PR #40, opened 2020-08-30
Consider path as symmetric on a given level of abstraction

cyrill-k:
I defined a path as symmetric only for a given level of abstraction and added IP routing symmetric paths as an example.

But I'm not convinced that covers all cases, as the order might not be relevant for certain service functions as Gorry pointed out.
What we could do is phrase the definition more generally:
"Two paths are symmetric if the path and its reverse path consist of the same path elements on the same level of abstraction, ~~but~~ **typically** in reverse order"

### DIFF
--- a/draft-irtf-panrg-path-properties.md
+++ b/draft-irtf-panrg-path-properties.md
@@ -196,7 +196,7 @@ Disjointness:
 
 Symmetric Path:
 : Two paths are symmetric if the path and its reverse path consist of the same path elements on the same level of abstraction, but in reverse order.
-A path which consists of layer-3 switches and links between them and a reverse path with the same path elements but in reverse order are considered "routing" symmetric, as the same path elements on the same level of abstraction (IP forwarding) are traversed in the opposite direction.
+For example, a path which consists of layer 3 switches and links between them and a reverse path with the same path elements but in reverse order are considered "routing" symmetric, as the same path elements on the same level of abstraction (IP forwarding) are traversed in the opposite direction.
 
 Path MTU:
 : The maximum size, in octets, of an IP packet that can be transmitted without fragmentation.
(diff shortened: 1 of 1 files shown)


draft-irtf-panrg-path-properties.md
```diff
@@ -195,7 +195,8 @@ Disjointness:
 : For a set of two paths or subpaths, the number of shared path elements can be a measure of intersection (e.g., Jaccard coefficient, which is the number of shared elements divided by the total number of elements). Conversely, the number of non-shared path elements can be a measure of disjointness (e.g., 1 - Jaccard coefficient). A multipath protocol might use disjointness as a metric to reduce the number of single points of failure.
 
 Symmetric Path:
-: Two paths are symmetric if a path and its reverse path consist of the same path elements, but in reverse order, i.e., the paths have a Jaccard coefficient of one.
+: Two paths are symmetric if the path and its reverse path consist of the same path elements on the same level of abstraction, but in reverse order.
+A path which consists of layer-3 switches and links between them and a reverse path with the same path elements but in reverse order are considered "routing" symmetric, as the same path elements on the same level of abstraction (IP forwarding) are traversed in the opposite direction.
```
theri, line 199:
```suggestion
For example, a path which consists of layer 3 switches and links between them and a reverse path with the same path elements but in reverse order are considered "routing" symmetric, as the same path elements on the same level of abstraction (IP forwarding) are traversed in the opposite direction.
```
This is an example of the previous definition, right? I think it's worth making this explicit.
Also, we already use "layer 2 domain" and "IP-layer" elsewhere in the draft, and I think it's good to try to be consistent. "IP-layer switch" sounds weird, therefore I'd go with "layer 3" (no hyphen) here.

cyrill-k:
I rephrased the service function example and added the textual fixes

theri (approved):
Thanks, looks good to me!

PR #41, opened 2020-08-30
Add table for different degrees of transparency

cyrill-k:
I added a table for different node behaviors (block, modify, ...) based on different inputs.
I'm not sure if this table is useful/consistent since different actions might have different inputs which cannot be represented in a reasonably small table.

An alternative approach instead of the table would be to add some text which describes the main takeaways from the table, i.e., node might block but not modify, node might modify headers but be transparent to payloads, ...

### DIFF
--- a/draft-irtf-panrg-path-properties.md
+++ b/draft-irtf-panrg-path-properties.md
@@ -185,19 +185,9 @@ Service function:
 : A service function that a path element applies to a flow, see {{RFC7665}}. Examples of abstract service functions include firewalls, Network Address Translation (NAT), and TCP optimizers. Some stateful service functions, such as NAT, require the same instance to be involved in both directions, i.e., on the path and the reverse path.
 
 Transparency:
-: A node perform an action transparently with respect to a protocol to a certain degree.
-The following table shows which action of a node (block packet, modify header, modify header+payload) is influenced by which inputs (while ignoring the not listed inputs).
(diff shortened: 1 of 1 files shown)


cyrill-k:
I changed it to a textual description.
What do you think?

draft-irtf-panrg-path-properties.md
```diff
@@ -185,8 +185,9 @@ Service function:
 : A service function that a path element applies to a flow, see {{RFC7665}}. Examples of abstract service functions include firewalls, Network Address Translation (NAT), and TCP optimizers. Some stateful service functions, such as NAT, require the same instance to be involved in both directions, i.e., on the path and the reverse path.
 
 Transparency:
-: A node is transparent with respect to a protocol if it does not modify headers of this protocol and it processes packets independently of protocol-specific meta-information.
-An IP router could be transparent for transport protocols such as TCP and UDP, in contrast to a NAT that actively modifies TCP and UDP header information.
+: We say that a node is transparent with respect to a protocol header, payload, or both for a specific action if the node performs this action independently of the given (meta-)information.
```
theri, line 188:
```suggestion
: A node is transparent with respect to a protocol header, payload, or both for a specific action if the node performs this action independently of the given (meta-)information.
```

theri (approved):
Looks good to me, modulo a stylistic change to one sentence.

PR #42, opened 2020-09-01
Quote path awareness definition and relationship to this draft

theri:
Reference the definition of path awareness in Section 1.1 of the Questions draft.
Also, make the initial statement less general - The current Internet does not support explicit discovery of path properties, but endpoints already do measure some properties of their paths (e.g., RTT measurements in a TCP stack).


cyrill-k:
This change looks good to me.

PR #43, opened 2020-09-01
Link to ALTO drafts

theri:
Highlight that our definition is compatible with ALTO by giving it as an example.
State that path properties can be made available as ALTO cost metrics.
Also, remove reference to an old paper of mine that has become irrelevant to this draft (IMO), as it is more about how to measure path properties, which is explicitly out of scope for this draft.


cyrill-k:
I added a small fix but I like the changes otherwise.

PR #44, opened 2021-02-12
Adjust definition to questions draft, rephrase different levels of abstraction

theri:
This PR does the following:

- Update the definition of "path aware networking" based on the current questions draft
- In the use cases section and the examples section, clarify how different entities have visibility of and control over different path elements and path properties. (Implicit: This is why the generic definition in this document becomes a specific definition in the context of a specific technology)
- In the Examples section, remove some text around examples that are not really path-aware networking
- Minor editorial fixes

Please let me know what you think.

### DIFF
--- a/draft-irtf-panrg-path-properties.md
+++ b/draft-irtf-panrg-path-properties.md
@@ -132,7 +132,7 @@ these entities may use this information to achieve different goals.
 This section lists several use cases for path properties.
 
 Note that this is not an exhaustive list, as with every new technology and protocol, novel use cases may emerge, and new path properties may become relevant.
-Moreover, different technologies may have visibility of and control over different path elements and path properties, and consider them on different levels of abstraction.
+Moreover, for any particular technology, entities may have visibility of and control over different path elements and path properties, and consider them on different levels of abstraction.
 Therefore, a new technology may implement an existing use case, but related to different path elements or on a new level of abstraction.
 
 ## Path Selection
--- a/draft-irtf-panrg-path-properties.md
+++ b/draft-irtf-panrg-path-properties.md
@@ -72,11 +72,12 @@ e.g., for selecting between paths or for invoking some of the provided services.
 # Introduction
 
 The current Internet architecture does not explicitly support endpoint discovery of forwarding paths through the network as well as the discovery of properties and services associated with these paths.
-Path-aware networking, as defined in Section 1.1 of {{I-D.irtf-panrg-questions}}, describes "endpoint discovery of the properties of paths they use for communication, and endpoint reaction to these properties that affects routing and/or transmission".
-This document defines such information as path properties, addressing the first of the questions in path-aware networking {{I-D.irtf-panrg-questions}}.
+Path-aware networking, as defined in Section 1.1 of {{I-D.irtf-panrg-questions}}, describes
+"endpoint discovery of the properties of paths they use for communication across an internetwork, and endpoint reaction to these properties that affects routing and/or data transfer".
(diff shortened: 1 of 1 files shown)


draft-irtf-panrg-path-properties.md
```diff
@@ -129,7 +130,10 @@ For example, a path property based on an approximate calculation may describe th
 When a path-aware network exposes path properties to hosts or other entities,
 these entities may use this information to achieve different goals.
 This section lists several use cases for path properties.
+
 Note that this is not an exhaustive list, as with every new technology and protocol, novel use cases may emerge, and new path properties may become relevant.
+Moreover, for any particular technology, entities may have visibility of and control over different path elements and path properties, and consider them on different levels of abstraction.
+Therefore, a new technology may implement an existing use case, but related to different path elements or on a new level of abstraction.
```
cyrill-k, line 136: I would prefer not to say a **new** level of abstraction, since this seems to indicate that our definition does not yet cover this level of abstraction:
"Moreover, for a particular technology, entities may have visibility of and control over different path elements and path properties, and consider them on different levels of abstraction.
Therefore, a new technology may implement an existing use case related to other path elements or on a different level of abstraction."

draft-irtf-panrg-path-properties.md
```diff
@@ -166,17 +170,15 @@ Another example is a connection which is composed of multiple streams where each
 
 This Section gives some examples of path properties which may be useful, e.g., for the use cases described in {{use-cases}}.
 
+Within the context of any particular technology, entities may or may not have insight into and be able to influence specific path elements and their path properties.
+Therefore, the path properties available in each such context may differ.
+For example, a host may have some visibility into path elements that are close, i.e., within the first few hops, or they may include path elements that are far away and/or on a different level of abstraction, e.g., the list of ASes traversed.
+The visibility of path elements and path properties to a specific entity may depend on factors such as the physical or network distance or the existence of trust or contractual relationships between the entity and the path element(s).
```
cyrill-k, line 176: Maybe we can rephrase this paragraph more concisely:
"Within the context of a particular technology, available path properties may differ as entities have insight into and are able to influence different path elements and path properties.
For example, a host may have some visibility into path elements that are close, i.e., within the first few hops, or they may include path elements that are far away and/or on a different level of abstraction, e.g., the list of ASes traversed.
Additionally, the visibility may depend on factors such as the physical or network distance or the existence of trust or contractual relationships between the entity and the path element(s)."
_should we also use **host** in the last sentence as in the sentence before?_

cyrill-k (changes requested):
I like the changes as they make the concept of visibility and abstraction more clear.

draft-irtf-panrg-path-properties.md
theri: Agreed! Fixed.

draft-irtf-panrg-path-properties.md
theri: Thanks, I used the first sentence you suggested.
Then, re-reading the second sentence, I rephrased it a bit, as I think its grammar was broken and I wanted more symmetry between the two options (low VS higher abstraction level and being close VS far away). Please let me know what you think.
For the last sentence, yes, let's stay with the example and use "host" here. But I intended this sentence to be a possible explanation for the sentence before, so it's not really "Additionally".

theri:
Thank you for the review! 
I updated the PR. Please let me know what you think.

draft-irtf-panrg-path-properties.md
cyrill-k: The text looks good now

theri:
Given that all previous comments are resolved, I'm merging this now.

PR #45, opened 2021-02-17
Technology-specific terminology

cyrill-k:
This text emphasizes that the terminology is intended to be general and applicable for a wide range of path aware technologies.

### DIFF
--- a/draft-irtf-panrg-path-properties.md
+++ b/draft-irtf-panrg-path-properties.md
@@ -124,6 +124,13 @@ Assessed property:
 : An approximate calculation or assessment of the value of a property. An assessed property includes the reliability of the calculation or assessment. The notion of reliability depends on the property.
 For example, a path property based on an approximate calculation may describe the expected median one-way latency of packets sent on a path within the next second, including the confidence level and interval. A non-numerical assessment may instead include the likelihood that the property holds.
 
+## Terminology usage for specific technologies
+
+The terminology described in this document is intended to be general and applicable to existing and future path-aware technologies.
+This document attempts to unify terminology which is used inconsistently between different path aware technologies.
+Using the terminology defined in this document, a path-aware technology can define its path elements on a specific level of abstraction.
+For example, a technology might define paths as sequence of IP routers as in source routing ({{?RFC1940}}), on different layers ({{?RFC1122}}), or as a collection of entities such as an AS or an ERO.
+
 # Use Cases for Path Properties {#use-cases}
 
 When a path-aware network exposes path properties to hosts or other entities,
--- a/draft-irtf-panrg-path-properties.md
+++ b/draft-irtf-panrg-path-properties.md
@@ -126,9 +126,8 @@ For example, a path property based on an approximate calculation may describe th
 
 ## Terminology usage for specific technologies
 
-The terminology described in this document is intended to be general and applicable to existing and future path-aware technologies.
-This document attempts to unify terminology which is used inconsistently between different path-aware technologies.
-Using the terminology defined in this document, a path-aware technology can define its path elements on a specific level of abstraction.
+The terminology defined in this document is intended to be general and applicable to existing and future path-aware technologies.
+Using this terminology, a path-aware technology can define and consider specific path elements and path properties on a specific level of abstraction.
 For example, a technology might define paths as sequence of IP routers as in source routing ({{?RFC1940}}), on different layers ({{?RFC1122}}), or as a collection of entities such as an AS or an ERO.
 
 # Use Cases for Path Properties {#use-cases}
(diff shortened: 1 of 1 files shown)


draft-irtf-panrg-path-properties.md
```diff
@@ -124,6 +124,13 @@ Assessed property:
 : An approximate calculation or assessment of the value of a property. An assessed property includes the reliability of the calculation or assessment. The notion of reliability depends on the property.
 For example, a path property based on an approximate calculation may describe the expected median one-way latency of packets sent on a path within the next second, including the confidence level and interval. A non-numerical assessment may instead include the likelihood that the property holds.
 
+## Terminology usage for specific technologies
+
+The terminology described in this document is intended to be general and applicable to existing and future path-aware technologies.
```
theri, line 129:
```suggestion
The terminology defined in this document is intended to be general and applicable to existing and future path-aware technologies.
```
I think "define terminology" is more fitting than "describe terminology" in this context. (Also, I'm getting 42.5k Google hits for "define terminology" and just 5.5k for "describe terminology" :smile:)

draft-irtf-panrg-path-properties.md
```diff
@@ -124,6 +124,13 @@ Assessed property:
...
+This document attempts to unify terminology which is used inconsistently between different path-aware technologies.
```
theri, line 130: I'm not sure about this sentence. Maybe we can delete it?

The introduction already gives the motivation for the common terminology:
"As terms related to paths have different meanings in different areas of networking, first, this document provides a common terminology to define paths, path elements, and path properties."

And I'm not sure if we really unify existing terminology - we did survey existing terminology, but "unify" sounds like a stronger implication that what the doc actually does, and I'm not sure if we even "attempt to unify".

If we want to add motivation to this section, how about adding the following to the first sentence above?
"[…] to provide a common language for the Path-Aware Networking Research Group."
Because I think that's actually the main motivation for having a generic common terminology that applies to different areas of networking.

draft-irtf-panrg-path-properties.md
```diff
@@ -124,6 +124,13 @@ Assessed property:
...
+Using the terminology defined in this document, a path-aware technology can define its path elements on a specific level of abstraction.
```
theri, line 131:
```suggestion
Using this terminology, a path-aware technology can define and consider specific path elements and path properties on a specific level of abstraction.
```
If the drop the second sentence, this sentence would follow after the first, so we can remove redundancy.
Also, added a "specific" to contrast with "generic".

draft-irtf-panrg-path-properties.md
```diff
@@ -124,6 +124,13 @@ Assessed property:
...
+For example, a technology might define paths as sequence of IP routers as in source routing ({{?RFC1940}}), on different layers ({{?RFC1122}}), or as a collection of entities such as an AS or an ERO.
```
theri, line 132:
```suggestion
For instance, a technology may define path elements as IP routers, e.g., in source routing ({{?RFC1940}}). Alternatively, it may consider path elements on a different layer of the Internet Architecture ({{?RFC1122}}) or independently, e.g., as a collection of entities, such as an AS or an ERO.
```

I think the original sentence listed different things that were not directly comparable, so I tried modifying it.
Also, as the previous sentence mentions "path elements" (and not "paths"), I think it's good to continue with path elements here.

theri (changes requested):
Thank you for taking a stab at this! I like the content overall and I made a few suggestions on wording.

draft-irtf-panrg-path-properties.md
cyrill-k: Agreed ;)

draft-irtf-panrg-path-properties.md
cyrill-k: I like your suggestion better as well, let's change it.

draft-irtf-panrg-path-properties.md
cyrill-k: That sounds good.

draft-irtf-panrg-path-properties.md
cyrill-k: > Also, as the previous sentence mentions "path elements" (and not "paths"), I think it's good to continue with path elements here.

Agreed

For me using "independently" in this context is a bit confusing (I guess the meaning is independent from specific Internet Architecture layers). Maybe we can rephrase it?

```suggestion
For instance, a technology may define path elements as IP routers, e.g., in source routing ({{?RFC1940}}). Alternatively, it may consider path elements on a different layer of the Internet Architecture ({{?RFC1122}}) or as a collection of entities not tied to a specific layer, such as an AS or an ERO.
```

draft-irtf-panrg-path-properties.md
cyrill-k:
```suggestion
```

theri (approved):
Thanks, this looks good to me now! Will merge and publish.

Issue #46, opened 2021-03-26
Clarify Node definition for microscopic context


theri:
From Jake's review, related to the "Node" definition in Section 2:

> Just to clarify: is for example a DSCP-selected queue in a switch a
> Node under this definition?  I think so, as it may have some specific
> relevant properties that are different from other queues, such as latency
> targets and capacity limits.
> 
> But as written the answer seems unclear to me, and I feel like it might
> cause confusion if used that way.  Would it be helpful to include
> something this as an example too?
> I think the AS example shows some helpful scope in the sense that this
> aggregates to a macroscopic context, but if I'm understanding correctly
> that it's intended to be generalizable like this, I think it might be
> helpful to clarify that depending on the properties and dynamics under
> discussion, it can also be useful to slice to a more microscopic context
> than a whole device or virtual device.

Yes, a DSCP-selected queue in a switch can be a Node.

To illustrate that a Node is not necessarily a physical or virtual machine, but can be defined in a more microscopic context, we have already included a service function as an example. To make it clearer, we can also add "a single queue" as another example for a (virtual) Node.

theri:
Comment from Med:

[Med] Another example we can add is the VRF or the more general concept of "VPN Node" defined in:

draft-ietf-opsawg-l3sm-l3nm-07:

   VPN node:  An abstraction that represents a set of policies applied
      on a PE and that belong to a single VPN service.  A VPN service
      involves one or more VPN nodes.  As it is an abstraction, the
      network controller will take on how to implement a VPN node.  For
      example, typically, in a BGP-based VPN, a VPN node could be mapped
      into a Virtual Routing and Forwarding (VRF).

PR #50, opened 2021-06-24
Add microscopic node example

theri:
Add microscopic node example as suggested by Jake in #46.
Considered adding "VPN node", but not sure if this is really necessary, so decided to not do so for now, to keep the definition short.


cyrill-k (commented):
I like this change to clarify the macroscopic concept, however I'd suggest to first resolve the entity definition issue and then incorporate this change.

theri:
Merged current "master" with the updated entity and node definition into this branch.
I think this can be merged now, right?

PR #51, opened 2021-06-24
Define "entity", be more deliberate in using entity VS node VS host, and traffic VS flows

theri:
When examining our use of "entity", I found that often we really mean anything that is either a node or measures path properties or processes them and potentially makes decisions based on them, e.g., path selection.
Therefore, I kept "entity" in most places, but added a definition as suggested by Med.

Also chose terms more deliberately in the "Path Selection" section - paths can be selected by any entity, not just the communicating node. And "traffic" is undefined so I substituted it by "flows" when we really just mean flows... or "data" when the packet has not actually been created yet (i.e., before a protocol has been selected).

### DIFF
--- a/draft-irtf-panrg-path-properties.md
+++ b/draft-irtf-panrg-path-properties.md
@@ -84,6 +84,9 @@ Note that this document does not assume that any of the listed path properties a
 
 # Terminology
 
+Entity:
+: A physical or virtual device or function which can process packets (i.e., a node or host), measure path properties, or access information about paths. With respect to a given communication (i.e., a flow), an entity may be on the data plane or control plane, and it may be on-path or off-path.
+
 Node:
 : An entity which processes packets, e.g., sends, receives, forwards, or modifies them. A node may be physical or virtual, e.g., a physical device or a service function provided as a virtual element. A node may also be the collection of multiple entities which, as a collection, processes packets, e.g., an entire Autonomous System (AS).
 
@@ -108,7 +111,7 @@ Subpath:
 : Given a path, a subpath is a sequence of adjacent path elements of this path.
 
 Flow:
-: An entity made of packets to which the traits of a path or set of subpaths may be applied in a functional sense. For example, a flow can consist of all packets sent within a TCP session with the same five-tuple between two hosts, or it can consist of all packets sent on the same physical link.
+: One or multiple packets to which the traits of a path or set of subpaths may be applied in a functional sense. For example, a flow can consist of all packets sent within a TCP session with the same five-tuple between two hosts, or it can consist of all packets sent on the same physical link.
 
 Property:
 : A trait of one or a sequence of path elements, or a trait of a flow with respect to one or a sequence of path elements. An example of a link property is the maximum data rate that can be sent over the link. An example of a node property is the administrative domain that the node belongs to. An example of a property of a flow with respect to a subpath is the aggregated one-way delay of the flow being sent from one node to another node over this subpath.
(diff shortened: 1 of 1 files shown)


draft-irtf-panrg-path-properties.md
```diff
@@ -143,26 +146,26 @@ Therefore, a new technology may implement an existing use case related to differ
 
 ## Path Selection
 
-Entities can choose what traffic to send over which path or subset of paths.
-A node might be able to select between a set of paths, either if there are several paths to the same destination (e.g., in case of a mobile device with two wireless interfaces, both providing a path), or if there are several destinations, and thus several paths, providing the same service (e.g., Application-Layer Traffic Optimization (ALTO) {{RFC5693}}, an application layer peer-to-peer protocol allowing hosts a better-than-random peer selection).
+Nodes may be able to send flows via one out of multiple possible paths, or via a subsets of paths, and an entity may be able to select which path(s) to use.
```
cyrill-k, line 149: Maybe we can rephrase this to clarify that these actions are not exclusive:
"Nodes may be able to send flows via one (or a subset) out of multiple possible paths and an entity may be able to influence the path selection."

draft-irtf-panrg-path-properties.md
```diff
@@ -84,6 +84,9 @@ Note that this document does not assume that any of the listed path properties a
 
 # Terminology
 
+Entity:
+: A physical or virtual device or function which can process packets (i.e., a node or host), measure path properties, or access information about paths. With respect to a given communication (i.e., a flow), an entity may be on the data plane or control plane, and it may be on-path or off-path.
+
 Node:
 : An entity which processes packets, e.g., sends, receives, forwards, or modifies them. A node may be physical or virtual, e.g., a physical device or a service function provided as a virtual element. A node may also be the collection of multiple entities which, as a collection, processes packets, e.g., an entire Autonomous System (AS).
```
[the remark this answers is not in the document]
cyrill-k: Actually, thinking about this discrepancy a bit more:
Since we added a clear definition of entity, and a node is defined as a collection of entities, it means not every node is an entity.
In the current draft, whenever we mention "entity" in the draft, we want to include any possible node and host.
Now, however, if we talk about an "entity" we only include nodes consisting of a single entity, which might not be what we want in many places of the text.

I think we could solve this by:
- changing the entity and node definition to ensure that every node is an entity (and include in the entity definition that an entity might consist of a collection of physical or virtual devices or functions).
- whenever we talk about an "entity" in the text and want to include the definition of a node, we make sure that we say "a node or an entity". However I think this is not a very elegant solution.

cyrill-k (changes requested):
I added some discussion in the terminology section about the new entity definition.

draft-irtf-panrg-path-properties.md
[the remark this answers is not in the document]
theri: Thanks for catching this.
I think the best way forward is to change the entity and node definition to ensure that every node is an entity.
I was actually intending to make "entities" a superset of "nodes", i.e., an entity is broader than a node, because we need nodes to define a path, and an entity can be basically anything and doesn't have to be on the path.

Changing the "Entity" definition now, and took a stab at simplifying the "Node" definition as well. I don't see the need to define a Node as a collection of entities now, as an entity can already be a collection. Do you agree?

draft-irtf-panrg-path-properties.md
```diff
@@ -84,6 +84,9 @@ Note that this document does not assume that any of the listed path properties a
 
 # Terminology
 
+Entity:
+: A physical or virtual device or function which can process packets (i.e., a node or host), measure path properties, or access information about paths. With respect to a given communication (i.e., a flow), an entity may be on the data plane or control plane, and it may be on-path or off-path.
```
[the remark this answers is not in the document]
theri: Good idea, will change.

draft-irtf-panrg-path-properties.md
theri: I'm not entirely convinced it said that the actions were exclusive, but I'm fine with shortening it a bit.
I like "influence", but I didn't want to explain path selection with itself, so I went for "influence the decision which path(s) to use". Does this work?

draft-irtf-panrg-path-properties.md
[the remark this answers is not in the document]
cyrill-k: Yes, if we phrase it in this way then it should be clear.

draft-irtf-panrg-path-properties.md
cyrill-k: Yes, that sounds good

draft-irtf-panrg-path-properties.md
```diff
@@ -85,10 +85,10 @@ Note that this document does not assume that any of the listed path properties a
...
+: A physical or virtual device or function, or a collection of devices or functions, which can, for example, process packets, measure path properties, or access information about paths. With respect to a given communication, an entity may be on the data plane or control plane, and it may be on-path or off-path.
 
 Node:
-: An entity which processes packets, e.g., sends, receives, forwards, or modifies them. A node may be physical or virtual, e.g., a physical device or a service function provided as a virtual element. A node may also be the collection of multiple entities which, as a collection, processes packets, e.g., an entire Autonomous System (AS).
+: An entity which processes packets, e.g., sends, receives, forwards, or modifies them. A node may be physical or virtual, e.g., a physical device or a service function provided as a virtual element. A node may also be an entity which, as a collection, processes packets, e.g., an entire Autonomous System (AS).
```
cyrill-k, line 91: We should maybe clarify what the collection consists of:
"An entity which processes packets, e.g., sends, receives, forwards, or modifies them. A node may be physical or virtual, e.g., a physical device or a service function provided as a virtual element. A node may also be an entity which consists of a collection of devices or functions, e.g., an entire Autonomous System (AS)."

theri: That works! Then it doesn't say anymore that the collection processes packets, but that's both already part of the definition and clear from the example.

theri:
I think we are aligned now, so I'll merge.

PR #52, opened 2021-06-24
Clarify context of Service Invocation use case, closes #49

### LINKED ISSUE #49: Service chaining vs. endpoint services
From [IETF 110 session](https://datatracker.ietf.org/doc/minutes-110-panrg/):

"Brian: good starting point, make sure service chaining vs. endpoint
    services is clear for service invocation (here we’re talking about service
    chaining). This doc has the problem of context overload for all the terms."

See also [recording](https://youtu.be/rgaknNfvtD4?t=1648).

theri:
Based on Brian's comment at IETF 110, setting the context explicitly: We're talking about Service Function Chaining.

### DIFF
--- a/draft-irtf-panrg-path-properties.md
+++ b/draft-irtf-panrg-path-properties.md
@@ -168,7 +168,7 @@ A video streaming application may choose an (initial) video quality based on the
 
 ## Service Invocation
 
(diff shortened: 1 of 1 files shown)


cyrill-k:
This clarification looks good to me.

PR #53, opened 2021-07-02
Slightly rephrase transparency definition and add more examples

cyrill-k:


draft-irtf-panrg-path-properties.md
```diff
@@ -196,9 +200,13 @@ Service function:
 : A service function that a path element applies to a flow, see {{RFC7665}}. Examples of abstract service functions include firewalls, Network Address Translation (NAT), and TCP optimizers. Some stateful service functions, such as NAT, need to observe the same flow in both directions, e.g., by being an element of both the path and the reverse path.
 
 Transparency:
-: A node is transparent with respect to a protocol header, payload, or both for a specific action if the node performs this action independently of the given (meta-)information.
+: A node is transparent to a flow with respect to some (meta-)information for an action if the node performs this action independently of the given (meta-)information.
```
theri, line 203: I found this sentence hard to parse.
How about something like:
```suggestion
: When a node performs an action on a flow, the node is transparent to the flow with respect to some (meta-)information if the node performs this action independently of the given (meta-)information.
```

draft-irtf-panrg-path-properties.md
```diff
@@ -196,9 +200,13 @@ Service function:
...
+(Meta-)information can for example be the existence of a protocol (header) in a packet or the content of a protocol header, payload, or both.
 Actions can for example be blocking packets or reading and modifying (other protocol) headers or payloads.
-An IP router could be transparent for transport protocol headers such as TCP and UDP, in contrast to a NAT that actively modifies TCP and UDP header information.
+An IP router could be transparent to transport protocol headers such as TCP/UDP but not transparent to IP headers as it actively modifies them.
+A TCP SYN flooding firewall that drops TCP SYN packets after reaching a limit is transparent to IP but not to TCP/UDP.
+And a NAT that actively modifies IP and TCP/UDP header information based on their content is not transparent to either IP or TCP/UDP headers.
```
theri, line 208: Nit, I don't like sentences starting with "And" :)
```suggestion
Finally, a NAT that actively modifies IP and TCP/UDP header information based on their content is not transparent to either IP or TCP/UDP headers.
```

draft-irtf-panrg-path-properties.md
```diff
@@ -196,9 +200,13 @@ Service function:
...
+Note that according to this definition, a node that modifies packets in accordance with the hosts, such as a transparent HTTP proxy, as defined in {{RFC2616}}, and a node listening and acting to implicit or explicit signals, see {{RFC8558}}, would not be considered transparent.
```
theri, line 209: Nit, I don't think we need the hypothetical here, and the node is reacting, right? Or maybe "acting on"?
```suggestion
Note that according to this definition, a node that modifies packets in accordance with the hosts, such as a transparent HTTP proxy, as defined in {{RFC2616}}, and a node listening and reacting to implicit or explicit signals, see {{RFC8558}}, are not considered transparent.
```

draft-irtf-panrg-path-properties.md
```diff
@@ -196,9 +200,13 @@ Service function:
...
 
 Transparency:
-: A node is transparent with respect to a protocol header, payload, or both for a specific action if the node performs this action independently of the given (meta-)information.
+: A node is transparent to a flow with respect to some (meta-)information for an action if the node performs this action independently of the given (meta-)information.
+(Meta-)information can for example be the existence of a protocol (header) in a packet or the content of a protocol header, payload, or both.
 Actions can for example be blocking packets or reading and modifying (other protocol) headers or payloads.
-An IP router could be transparent for transport protocol headers such as TCP and UDP, in contrast to a NAT that actively modifies TCP and UDP header information.
+An IP router could be transparent to transport protocol headers such as TCP/UDP but not transparent to IP headers as it actively modifies them.
+A TCP SYN flooding firewall that drops TCP SYN packets after reaching a limit is transparent to IP but not to TCP/UDP.
```
theri, line 207: If the firewall indeed does not look at the IP addresses at all, it would help to clarify the example, e.g.:
```suggestion
A TCP SYN flooding firewall that drops TCP SYN packets after having seen a certain number of packets with TCP SYN, regardless of IP address, is transparent to IP but not to TCP/UDP.
```

But it doesn't sound like a good idea to implement it this way, as that means you could DoS a network by SYN-flooding it. 
So perhaps this firewall should count TCP SYN packets from a particular IP address or for a combination of source and destination IP addresses, in which case it would not be transparent to either TCP or IP.

What do you think?

theri (changes requested):
Thanks for these changes, I like them, and I have a few proposed tweaks.

I'm not sure what to do about the "endpoint transparency" ("host transparency"?), or the explicit communication with a host.
If the node has communicated with the host, say, out-of-band, and then modifies the packet based on that, can that be just another piece of meta-information?

I'm not sure we should attempt to capture a broader concept of "endpoint transparency", e.g., whether an endpoint is aware of a particular node, and whether it communicates with the node explicitly. Maybe we can get away with just making this awareness or communication be "meta-information", which we don't define further?

draft-irtf-panrg-path-properties.md
cyrill-k: I wanted to have an example where a node is transparent to IP but not to TCP/UDP to show that such a combination is also possible.
But I agree that the example does not make much sense in practice ^^

Maybe we could use the example of a firewall that only allows outgoing TCP connection, i.e., blocks all incoming TCP SYN packets regardless of the IP address?
That seems more reasonable.

cyrill-k:
> I'm not sure what to do about the "endpoint transparency" ("host transparency"?), or the explicit communication with a host.
If the node has communicated with the host, say, out-of-band, and then modifies the packet based on that, can that be just another piece of meta-information?

I guess that is true, however even in such a case, the relevant piece of meta-information is how to **identify** the flow and associate it to host, with which the node communicated to perform a specific action.
And the identification relies on the existence and content of protocol headers and payloads, which we cover with our definition.
So, for example to model RFC8558 style interactions between hosts and on-path nodes using our path terminology, we could say that the explicit communication with a host changes the transparency property of the node.

> I'm not sure we should attempt to capture a broader concept of "endpoint transparency", e.g., whether an endpoint is aware of a particular node, and whether it communicates with the node explicitly. Maybe we can get away with just making this awareness or communication be "meta-information", which we don't define further?

I would suggest to keep the definition as it is right now and keep the definition of meta-information open-ended.

draft-irtf-panrg-path-properties.md
theri: Yep, sounds like this is a better example.

theri:
I agree with your reasoning regarding the "endpoint transparency" - let's keep the definition as-is.

So I think this PR can be merged after changing the firewall example. :slightly_smiling_face:

cyrill-k:
I changed the firewall example.
If you agree with the changes, you can merge the PR.

draft-irtf-panrg-path-properties.md
```diff
@@ -196,9 +200,13 @@ Service function:
...
 
 Transparency:
-: A node is transparent with respect to a protocol header, payload, or both for a specific action if the node performs this action independently of the given (meta-)information.
+: When a node performs an action on a flow, the node is transparent to the flow with respect to some (meta-)information if the node performs this action independently of the given (meta-)information.
+(Meta-)information can for example be the existence of a protocol (header) in a packet or the content of a protocol header, payload, or both.
 Actions can for example be blocking packets or reading and modifying (other protocol) headers or payloads.
-An IP router could be transparent for transport protocol headers such as TCP and UDP, in contrast to a NAT that actively modifies TCP and UDP header information.
+An IP router could be transparent to transport protocol headers such as TCP/UDP but not transparent to IP headers as it actively modifies them.
```
theri, line 206: Just reading the whole thing again: Why does an IP router actively modify IP headers, whereby the text reads as if that's necessarily always the case?
Wouldn't the router just look up the destination address in its forwarding table and then forward the packet, modifying the Ethernet header for sure, but not the IP header?
Sorry, am I just missing something right now? :sweat_smile:

theri (commented):
The firewall example looks good, but I just tripped over the IP router example. Sorry, didn't see this earlier. And it might just be me missing something obvious...?

draft-irtf-panrg-path-properties.md
theri: I guess the IP router reads these headers for sure, so the example wouldn't have to change much to say this.

cyrill-k: I was thinking that the IP router changes certain IP header fields (e.g., TTL, IP checksum), however maybe we can make a better example by considering the action of "forwarding on a certain interface".
What do you think?

```suggestion
An IP router could be transparent to transport protocol headers such as TCP/UDP but not transparent to IP headers since its forwarding behavior depends on the IP headers.
```

theri: I like your suggestion. And you're right, "modify IP header" doesn't necessarily mean addresses... but still I think your proposed wording is clearer.

theri:
Looks good to me now, so I'm merging it.

PR #54, opened 2021-07-02
Add point-multipoint example to protocol feature property

cyrill-k:


theri:
Thanks for taking a stab, but I don't think it really fits under "protocol features" - these were intended to be features of a single given protocol, which can be switched on or off. Replication efficiency, as outlined by Jake, appears to be more about differences between multiple solutions which all share the concept of replicating packets, but are otherwise very different protocols, and may even be tied to a specific physical link type.

In my reply I had asked about this property's relationship to a use case, and I have not gotten a reply.
I am still not seeing a use case myself - I don't think it can be used for Path Selection, maybe it could have some relationship to Protocol Selection but it's not obvious to me and would need an example, and I don't think there's a relationship to Service Invocation.
So, we could leave it out for now and maybe ask Jake again what use case he had in mind?

Issue #55, opened 2021-10-12
Define "data plane" and "control plane"?


renghardt:
In a discussion around draft-irtf-panrg-questions-10, Spencer suggested that maybe the path properties draft should define "data plane" and "control plane", as it uses these terms in the definition of "entity", and as other PANRG drafts also use these terms.

The path propertiest draft uses these terms only to say the following:
>  an entity may be on the data plane or control plane
Not sure if that makes it critical to define these terms.

The discussion around draft-irtf-panrg-questions-10 has now resulted in using an alternative term for that draft.

If we decide that the path properties draft needs to define "control plane" and "data plane", here's some first thougths:

The path properties draft already cites [RFC 7665](https://www.rfc-editor.org/rfc/rfc7665.html#section-5.2), which talks about an SFC Control plane and somewhat outlines what it is:
> The SFC control plane is responsible for constructing SFPs,
> translating SFCs to forwarding paths, and propagating path
> information to participating nodes to achieve requisite forwarding
> behavior to construct the service overlay.

But RFC 7665 states that "the detailed definition of the SFC control plane is outside the scope of this document." I haven't seen other definitions in an RFC, but that doesn't mean there aren't any. Perhaps the SFC Control plane is too specific for the path properties draft anyways.

Doing a web search, I found more general definitions of these terms, e.g., [here](https://www.cloudflare.com/learning/network-layer/what-is-the-control-plane/):
> The control plane is the part of a network that controls how data packets are forwarded — meaning how data is sent from one place to another. The process of creating a routing table, for example, is considered part of the control plane.

> In contrast to the control plane, which determines how packets should be forwarded, the data plane actually forwards the packets. The data plane is also called the forwarding plane.

Any opinion?

Issue #47, opened 2021-06-22
Clarify "entity"


theri:
From Jake's review:

> 
> I think there's a lot of use of the word "entity" when a narrower "Host"
> or "Node" would be better.
>
> I'm getting a vague impression that in some situations, like when
> information is exposed to an entity that can use it to achieve goals,
> like in "3. Use Cases for Path Properties", "entity" is intended to
> capture nodes in general plus things like traffic engineering
> controllers that might have no part in a traffic path, whereas in
> other cases I think it refers only to a host (like in "3.2 Protocol
> Selection"--is there any non-host that can select a protocol for its
> data traffic?  When it's sending data traffic it's acting as a host,
> right?), and it would be nicer if I wasn't guessing at the intended
> scope.
>
> I think there's a few ways to handle it, but my first suggestion would
> be to use "entity" or "entities" as-is in the definitions, and try to
> excise those terms from the rest of the doc, and if there are cases that
> can't use a word that's in the terminology section (or combinations like
> "nodes or links") as a substitution without losing the intended meaning,
> it suggests another term might be needed.

Yes, "entity" is intended to be broader than just nodes on the path and can include off-path devices like traffic engineering controllers. It doesn't have to be a physical device either, it could also be a service running "in the cloud".

I agree it would be good to substitute "entity" with a more specific term in many instances later in the document.

However, I am not sure what a good term for these is, as I'm not sure I want to overload "Node". But maybe it actually is a Node, which may be on-path or off-path, and maybe in some cases it has to be on-path? (Though I imagine there could be cases where an off-path Node actually selects a protocol that some on-path Node then uses, or at least attempts to use.)

theri:
Comment from Med:

[Med] Agree. That can be captured by making sure to convey the message that these entities may not belong necessary to the data plane, but to the control plane.

theri:
Feedback from the IETF 111 PANRG session:
We need to be more specific. 

Richard basically said it only makes sense to have the entity term if all entities have something in common, otherwise it's too abstract.

This makes me think that perhaps our list of examples of what an entity could be doing is not enough. But maybe we can distill some commonalities from that list and/or our use cases? Maybe something like, an entity observes/measures path properties (or is somehow aware of them), and it can influence the path taken by flows (e.g., Path Selection) or the flow itself (e.g., Protocol Selection and Service invocation).

PR #56, opened 2021-10-19
Expand and clarify entity definition

renghardt:
This PR addresses #47 and #55 by explicitly stating a common aspect of all entities (in the scope of this document): They play a role related to path-aware networking for a particular path and flow.
It then illustrates these roles based on the clarified "data plane" and "control plane".

Not sure if "what may be called [data|control] plane" is the right phrasing yet, but I didn't want to imply that one *has* to use these terms, or that the entity's role always has to align with any possible definition of these terms.

Please let me know what you think.

### DIFF
--- a/draft-irtf-panrg-path-properties.md
+++ b/draft-irtf-panrg-path-properties.md
@@ -89,12 +89,13 @@ Note that this document does not assume that any of the listed path properties a
 # Terminology
 
 Entity:
-: A physical or virtual device or function, or a collection of devices or functions, which plays a role related to path-aware networking for a particular path and flow.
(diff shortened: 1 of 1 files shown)


cyrill-k:
I like the changes and I think its good to keep it as general as possible.
I changed a few things:
- Change "control aspects" -> "influence aspects" since the actual path control may be done by the endpoint not the control plane (although the name *control* plane seems to imply control so we can also keep "control aspects" ;D )
- Change node definition to be an **on-path** entity because we want to actually use the abstract entity definitions (#47).
- Also, I changed some occurrences of *node* to *entity* in the security considerations since it seems to be more general.

What do you think?

Issue #62, opened 2022-02-04
Definitions depend on context


theri:
At the [IETF 112 PANRG session](https://youtu.be/m9uA2XKlmeU?t=516), Sabine Randriamasy brought up the point that a definition can be relative to different scopes, e.g., an endpoint can mean different things in routing VS transport, or for flows VS applications.

While Section 2.1 of the path properties draft already mentions that the terminology in the document is applicable to different technologies, maybe it should mention the different contexts as well. Or this could be part of the introduction.

Quoting from two emails by @cyrill-k:
> "My suggestion is to mention in the introduction that even for a single PAN technology, there might be different contexts (e.g., data- and control-plane) where the entities might have different behavior and properties."
> 
> "[…] maybe we could emphasize that even for a single PAN technology, different contexts, e.g., data- and control-plane, can use a different terminology in terms of nodes, links, and thus paths."

cyrill-k:
I think it should be sufficient to add a few sentences to Section 2.1 to address the issue of contexts for specific technologies.
I would not put it in the introduction, since in my opinion it is not necessary to think about different contexts to understand the terminology.

PR #63, opened 2022-02-22
Add brief discussion of contexts in the draft

cyrill-k:
I added two sentences about contexts.
See #62

### DIFF
--- a/draft-irtf-panrg-path-properties.md
+++ b/draft-irtf-panrg-path-properties.md
@@ -141,7 +141,7 @@ The terminology defined in this document is intended to be general and applicabl
 Using this terminology, a path-aware technology can define and consider specific path elements and path properties on a specific level of abstraction.
 For instance, a technology may define path elements as IP routers, e.g., in source routing ({{?RFC1940}}). Alternatively, it may consider path elements on a different layer of the Internet Architecture ({{?RFC1122}}) or as a collection of entities not tied to a specific layer, such as an AS or an ERO.
 Even within a single path-aware technology, specific definitions might differ depending on the context in which they are used.
-For example, the start or end node might be the communicating host in the data-plane context, and the AS of the respective host in the control-plane context.
+For example, the endpoints might be the communicating hosts in the context of the transport layer, ASes that contain the hosts in the context of routing, or specific applications in the context of the application layer.
 
 # Use Cases for Path Properties {#use-cases}
 
--- a/draft-irtf-panrg-path-properties.md
+++ b/draft-irtf-panrg-path-properties.md
@@ -140,6 +140,8 @@ For example, a path property based on an approximate calculation may describe th
 The terminology defined in this document is intended to be general and applicable to existing and future path-aware technologies.
 Using this terminology, a path-aware technology can define and consider specific path elements and path properties on a specific level of abstraction.
 For instance, a technology may define path elements as IP routers, e.g., in source routing ({{?RFC1940}}). Alternatively, it may consider path elements on a different layer of the Internet Architecture ({{?RFC1122}}) or as a collection of entities not tied to a specific layer, such as an AS or an ERO.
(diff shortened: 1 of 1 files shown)


draft-irtf-panrg-path-properties.md
```diff
@@ -140,6 +140,8 @@ For example, a path property based on an approximate calculation may describe th
 The terminology defined in this document is intended to be general and applicable to existing and future path-aware technologies.
 Using this terminology, a path-aware technology can define and consider specific path elements and path properties on a specific level of abstraction.
 For instance, a technology may define path elements as IP routers, e.g., in source routing ({{?RFC1940}}). Alternatively, it may consider path elements on a different layer of the Internet Architecture ({{?RFC1122}}) or as a collection of entities not tied to a specific layer, such as an AS or an ERO.
+Even within a single path-aware technology, terminology might differ depending on the context in which it is used.
```
theri, line 143: Is it really the terminology that differs, or is it the specific definition you get when using this terminology, i.e., what specific component gets defined as a node/path/endpoint? (At least that's what the prior few sentences suggest, as I read them.)

I'm wondering if this should say:
```suggestion
Even within a single path-aware technology, specific definitions might differ depending on the context in which they are used.
```

draft-irtf-panrg-path-properties.md
```diff
@@ -140,6 +140,8 @@ For example, a path property based on an approximate calculation may describe th
...
+For example, the start or end node might be the communicating host in the data-plane context, and the AS of the respective host in the control-plane context.
```
theri, line 144: If our endpoint definition lands, we can change this to "endpoint".

However, I'm not quite sure what "the data-plane context" and "the control-plane context" actually is. We do define data plane and control plane as part of the entity definition, but I still don't think these are necessarily clearly defined contexts. At least in my mind, a "control-plane context" could just as easily refer to routing between individual hosts and not just be on an AS level.

An easy fix would be to make it sound less like there is *the* single clearly defined "control plane context":

```suggestion
For example, the endpoints might be the communicating hosts in a data-plane context, and the AS of the respective hosts in a control-plane context.
```

Or maybe we can change the example:

```suggestion
For example, the endpoints might be the communicating hosts in the context of the transport layer, ASes that contain the hosts in the context of routing, or specific applications in the context of the application layer.
```

However, at this point, it feels like we're almost replicating the part of the Path definition which talks about levels of abstraction:

> Path: […] Different entities may have different visibility of a path and/or treat path elements at different levels of abstraction. For example, a path may be given as a sequence of physical nodes and the links connecting these nodes, or it may be given as a sequence of logical nodes such as a sequence of ASes or an Explicit Route Object (ERO). Similarly, the representation of a path and its properties, as it is known to a specific entity, may be more complex and include details about the physical layer technology, or it may be more abstract and only consist of a specific source and destination which is known to be reachable from that source.

But maybe that's fine, as the point is important enough that it bears repeating, as we keep getting comments about it...? :slightly_smiling_face: 
(And no, I don't want this document to define "transport layer" or "routing"...)

theri (changes requested):
Thanks for the suggestion, I agree that a brief addition here is enough.
However, I do have quite a few thoughts on the specifics.

draft-irtf-panrg-path-properties.md
cyrill-k: An advantage of the second proposal is that it shows that an endpoint can also be a virtual entity, such as an application, on a host (not only the host itself). Also, it does not suffer from the fact that a reader may think, there are only two contexts, i.e., data-plane vs control-plane.

I would suggest to use the second proposal (even if the text is somewhat redundant, in my opinion it still helps with understanding).

theri: Ok, sounds good! Let's go with the second proposal then.

PR #65, opened 2022-02-24
Add "Discussion venues" section

theri:
It took me a while to figure out, but this PR adds a "Discussion venues" section to the draft, pointing to the Github repo and mailing list.

For future reference, what I recall doing was:

1. Copy the entire repo to a new directory (while already being on this branch)
2. `cd lib && git pull` to update the template repo
3. `make`, and then switch to using a "main" branch as advised by the output
4. Make sure I have no uncommited changes and that I am on master (I don't understand why being on master is a requirement)
5. `make -f lib/setup.mk` (which creates a commit)
6. Manually edit .note.xml to actually contain the correct URL and a pointer to the mailing list, similar to [Spencer's example](https://github.com/SpencerDawkins/sdp-rtp-quic-questions/blob/main/.note.xml)
7. Commit the changed .note.mk to "master". (Do not push)
8. Switch back to my branch and cherry-pick both commits from "master"
9. `make clean && make` -> Finally, the "Discussion venues" section shows up in the generated .txt and .html file
10. Commit and push the changes to this branch

Maybe there was an easier way to "update" a draft that was started before this feature existed, but I did not find it.


martinthomson:
I'm sorry you had to go through all of that.  I didn't really realize just how awkward this experience might be.

Conveniently, this draft uses krandowm-rfc, so `make update-venue` likely would have worked for you, with a fair bit less effort involved.

theri:
Thanks, I had no idea `make update-venue` existed, and I wasn't aware of the "make update-* features" in general. That would be a good addition to the documentation for the tools. :slightly_smiling_face: 
(I found that the CODEOWNERS file does say that it can be regenerated with `make update-codeowners` - I wonder if a similar comment could be added to other places where things can be regenerated. In case of the .note.mk, I had no file to start with - but I did look at Spencer's file and wondered how he generated it.)

martinthomson:
https://github.com/martinthomson/i-d-template/blob/main/doc/UPDATE.md is a start on documenting that.

Issue #58, opened 2021-10-29
Add "routing domain identifier"


SpencerDawkins:
In PANRG discussion about draft-irtf-panrg-questions, we tripped over that draft using the term "control plane address", with no definition, either in draft-irtf-panrg-questions or in the path-properties draft. 

Based on discussion on the PANRG mailing list, we seem to be changing that to "routing domain identifier" (discussion thread is at https://mailarchive.ietf.org/arch/msg/panrg/BzPUTAZo0apm1HQ-wTE7xIq6g30/).

Should this term be added to "path properties", since it's being used in our research group documents? 

(This is probably also a good question for @britram and @furry13}

theri:
Following the discussion at the [IETF 112 PANRG session](https://youtu.be/m9uA2XKlmeU?t=516), it sounds to me like we should add this definition if it doesn't have an obvious, widely understood meaning.

It seems like "routing domain" does have a definition in [RFC1237](https://www.rfc-editor.org/in-notes/rfc1237.txt), it is "a collection of networked systems that operate common routing protocols and are under the control of a single administration".
Then, [RFC 1629](https://datatracker.ietf.org/doc/html/rfc1629) defines the OSI Interdomain Routing Protocol (IDRP), which has the concept of a Routing Domain Identifier (RDI), and [RFC1930](https://www.rfc-editor.org/in-notes/rfc1930.txt) says "It should be noted that the IDRP equivalent of an AS is the RDI, or Routing Domain Identifier.". 
These definitions seem kind of historic, and I haven't found any more current RFCs.

But I did find:
- a random website claiming [a "routing domain" is an AS](https://www.netinbag.com/en/internet/what-is-a-routing-domain.html), so a routing domain identifier would be an ASN.
- F5 defining a [route domain](https://techdocs.f5.com/kb/en-us/products/big-ip_ltm/manuals/product/tmos-routing-administration-11-6-0/8.html) as a configuration object that isolates network traffic for a particular application on the network. I'm not sure if this is the same as a routing domain, but the latter term does appear on that page, too, and it looks like it's used as a synonym.
- A [patent application by Nicira](https://patents.justia.com/patent/20200382407), which talks about ways to assign a routing domain identifier to one or more logical switches, but I'm not seeing a proper definition in there.

[...]

PR #66, opened 2022-02-28
Rephrase administrative domain and add routing domain identifier definition

cyrill-k:
I rephrased the definition of administrative domain and added a definition for routing domain identifier as discussed in #58.

I'm wondering if it may be more elegant to define "routing domain" and then only mention its identifier in this definition:
"A routing domain is an administrative domain, in which all path elements use a common routing protocol to communicate with each other.
The routing domain identifier is a unique identifier of a routing domain.
An example of a routing domain identifier is the globally unique autonomous system number (ASN) as defined in {{RFC1930}}."

What do you think?

### DIFF
--- a/draft-irtf-panrg-path-properties.md
+++ b/draft-irtf-panrg-path-properties.md
@@ -224,7 +224,7 @@ Examples of administrative domains are an IGP area, an AS, or a service provider
 
 Routing Domain Identifier:
 : An identifier indicating the routing domain of a path element.
-All path elements with the same routing domain identifier are in the same administrative domain and use a common routing protocol to communicate with each other.
+Path elements in the same routing domain are in the same administrative domain and use a common routing protocol to communicate with each other.
 An example of a routing domain identifier is the globally unique autonomous system number (ASN) as defined in {{RFC1930}}.
 
 Disjointness:
--- a/draft-irtf-panrg-path-properties.md
+++ b/draft-irtf-panrg-path-properties.md
@@ -219,7 +219,7 @@ Finally, a NAT that actively modifies IP and TCP/UDP headers based on their cont
 Note that according to this definition, a node that modifies packets in accordance with the hosts, such as a transparent HTTP proxy, as defined in {{RFC2616}}, and a node listening and reacting to implicit or explicit signals, see {{RFC8558}}, are not considered transparent.
 
 Administrative Domain:
-: An individual or an organization that owns a path element (or several path elements).
+: The identity of an individual or an organization that owns a path element (or several path elements).
 Examples of administrative domains are an IGP area, an AS, or a service provider network.
 
 Routing Domain Identifier:
--- a/draft-irtf-panrg-path-properties.md
+++ b/draft-irtf-panrg-path-properties.md
@@ -39,6 +39,8 @@ informative:
 
     I-D.ietf-alto-performance-metrics:
 
+    RFC1930:
+
     RFC2616:
 
     RFC3357:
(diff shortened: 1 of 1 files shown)


draft-irtf-panrg-path-properties.md
```diff
@@ -217,7 +219,13 @@ Finally, a NAT that actively modifies IP and TCP/UDP headers based on their cont
 Note that according to this definition, a node that modifies packets in accordance with the hosts, such as a transparent HTTP proxy, as defined in {{RFC2616}}, and a node listening and reacting to implicit or explicit signals, see {{RFC8558}}, are not considered transparent.
 
 Administrative Domain:
-: The administrative domain, e.g., the IGP area, AS, or Service provider network to which a path element belongs.
+: An individual or an organization that owns a path element (or several path elements).
+Examples of administrative domains are an IGP area, an AS, or a service provider network.
```
theri, line 223: Thinking about the difference between "Administrative Domain" and "Routing Domain Identifier":
Maybe the example of the administrative domain would be an identifier for the IGP area, owner of the AS, or service provider network?

Of course, we could just make this property an "Administrative domain idenfitier".

draft-irtf-panrg-path-properties.md
```diff
@@ -217,7 +219,13 @@ Finally, a NAT that actively modifies IP and TCP/UDP headers based on their cont
...
+
+Routing Domain Identifier:
+: An identifier indicating the routing domain of a path element.
+All path elements with the same routing domain identifier are in the same administrative domain and use a common routing protocol to communicate with each other.
```
theri, line 227: I think here it would be fine to say the following, to make it more of a definition (i.e., these are necessary requirements for path elements to be in the same routing domain, instead of merely statements that are true for some path elements in the same routing domain).
And I think we can still put it like this, even if we keep this an "identifier".

```suggestion
Path elements in the same routing domain are in the same administrative domain and use a common routing protocol to communicate with each other.
```

theri (changes requested):
Thanks for the PR!

I intentionally wanted this to be "routing domain identifier" instead of just "routing domain" because these are examples of path properties (as per the section heading).
A path property, as defined, is a trait of one or multiple path elements. If I think of "annotating" one or more path elements with a path property, I would annotate them with a "routing domain identifier", I don't think I could annotate them with a "routing domain". If it's a routing domain, not identifier, to me it sounds like we're talking about some abstract notion of "grouping path elements" as a concept separate from path properties, instead of simply giving examples of path properties.
In other words, I would like these to clearly be examples of a path property that can be a trait of one or more path elements, instead of trying to define some concept of aggregations of path properties, which I think would confuse people.

Does that make sense to you?
Maybe the distinction is all in my head... :sweat_smile:

draft-irtf-panrg-path-properties.md
cyrill-k: I don't think there is a common type of identifiers for administrative domains. This is unlike the routing domain identifier property, which has a clearly defined semantics that is used by all routing domains to coordinate the routing between them.
Administrative domain seems more abstract and may not always have an identifier. However, I suppose it must have an identity to be a useful property ;D

I'm not sure how to include the word identity though...

cyrill-k: Do you have any suggestions how to integrate it?

theri, line 223: (Responding to the above question, but having to "comment" on two lines to make the suggestion work)

Good point that identifiers for administrative domain are less clearly defined. Perhaps we can leave this as "Administrative domain", and then defining an administrative domain implies providing some kind of identifier or identity that refers to the individual or organization that owns the path element(s) - at least that's the only way I see how one would do that.

If we want to include the word "identity", how about:
```suggestion
: The identity of an individual or an organization that owns a path element (or several path elements).
Examples of administrative domains are an IGP area, an AS, or a service provider network.
```

theri (approved):
I think this can now be merged as-is, or we can add the "identity".

PR #75, opened 2022-08-05
First stab at text to explain that some properties are very dynamic

theri:
I took a first stab at this text. Please let me know what you think!


cyrill-k:
I like the text and I think it clearly illustrates the problem.
However, we now have this discussion of potential problems caused by dynamic observed properties (and path selection based on such properties) in four places (here [middle of 4], first paragraph of 3.1, second paragraph of 3.1, and second paragraph of 4).

Should we consider merging some of them?
Maybe we could merge this text and the previous one (second paragraph of 4) and the two occurrences in section 3.1.

What do you think?

theri:
Oh, right. I merged the two paragraphs in Section 4 (and also changed some text to use "observed" and "assessed", as that's the terms we actually define).
We should merge the text in Section 3.1 as well, but I'll hold off on that here, as your other PR current reworks that section, and could incorporate that change as well.

cyrill-k:
This now looks good to me.
I agree, if necessary let's merge the text in 3.1 in the other PR (#77)

Issue #70, opened 2022-04-15
Try to make path definition clearer/more readable


theri:
Via @SpencerDawkins in Last Call:

> In the definition of “Path”, 
>
> For multicast or broadcast, a packet may be sent by one node and received by multiple nodes. In this case, the packet is sent over multiple paths at once, one path for each combination of sending and receiving node; these paths do not have to be disjoint.
>
> I think the point this text is making is that if a sending node performs broadcast or multicast over multiple paths, that’s not relevant to the definition of a path in this document - each one of the paths from sender to receiver(s) is just another path, according to this definition. 
>
> Is that correct? If so, perhaps this could be a second paragraph on its own within the definition, to be clearer. 

Yes, I would say that this is correct.

In general, I think the Path definition is quite long and includes lots of aspects. I'll see what we can do to make it clearer.

cyrill-k:
One easy fix could be to make the path definition minimal, i.e., "A sequence of adjacent path elements over which a packet can be transmitted, starting and ending with a node. A path is unidirectional.".
And then have the rest of the explanations in a separate paragraph below.

I think we can maybe rearrange some of the text to make the text clearer.
I can do a first pass for the rephrasing and make a PR.

PR #76, opened 2022-08-11
First draft for clarifying the definition of path

cyrill-k:
Rewrote the path definition to be more concise and clear (see #70).
Also, added paragraphs to separate the actual definition from the additional comments.


theri:
I like the restructuring, but I don't like how this renders currently.
The definition is only the first sentence, and then the other paragraphs are just sitting there unindented, between two definitions, and it's not clear that they belong to the previous definition (i.e., "Path").
Is it possible to have these additional paragraphs with newlines, but have them indented under the "Path" definition?

cyrill-k:
I fixed the indentation.
What do you think?

theri:
Nice, this looks good to me, thanks.
Merging.

PR #77, opened 2022-08-11
Add target property definition (fixes #69)

### LINKED ISSUE #69: Consider adding "Target property"
Via Last Call comment by @boucadair:

(1) Several application/services manipulate not only observed/assessed values, but can also set objectives/targets. I suggest we provision a new entry to cover that: 

NEW:
   Target property:  An objective that is set for a property over a path,
      subpath, or network.  For example, the maximum one-way delay to be 
      observed by packets transmitted from a node. Such properties can be
      captured on service requests (e.g., Section 4.1.1.1 of [I-D.ietf-teas-ietf-network-slices]).

cyrill-k:
Added a new definition Target property and rewrote section 3.1 to refer to this new term (see #69).
Also added a comparison to SLE, SLI, and SLOs with a reference to the IETF network slices draft.

### DIFF
--- a/draft-irtf-panrg-path-properties.md
+++ b/draft-irtf-panrg-path-properties.md
@@ -140,7 +140,7 @@ For example, a path property based on an approximate calculation may describe th
 
 Target property:
 : An objective that is set for a property over a path element, subpath, or path.
-Note that a target property can be set for observable properties, such as one-way delay, but also for properties that are not (easily) observable, such as inclusion of certain nodes on a path.
+Note that a target property can be set for observed properties, such as one-way delay, but also for properties that cannot be observed by the entity setting the target, such as inclusion of certain nodes on a path.
 
 ## Terminology usage for specific technologies
 
--- a/draft-irtf-panrg-path-properties.md
+++ b/draft-irtf-panrg-path-properties.md
@@ -164,8 +164,6 @@ Therefore, a new technology may implement an existing use case related to differ
 
 Nodes may be able to send flows via one (or a subset) out of multiple possible paths, and an entity may be able to influence the decision which path(s) to use.
 Path Selection may be feasible if there are several paths to the same destination (e.g., in case of a mobile device with two wireless interfaces, both providing a path), or if there are several destinations, and thus several paths, providing the same service (e.g., Application-Layer Traffic Optimization (ALTO) {{RFC5693}}, an application layer peer-to-peer protocol allowing endpoints a better-than-random peer selection).
-Care needs to be taken when selecting paths based on observed path properties, as path properties that were previously measured may not be helpful in predicting current or future path properties and such path selection may lead to unintended feedback loops.
-
 Entities may select their paths to fulfill a specific goal by specifying target properties for their selected paths, e.g., related to performance or security.
 
 Target properties relating to network performance typically refer to observed properties, such as One-Way Delay, One-Way Packet Loss, and Link Capacity.
@@ -176,7 +174,8 @@ As an example path selection strategy, for sending a small delay-sensitive query
 It is also possible for an entity to set target properties which it cannot (directly) observe, similar to service level expectations (SLEs) for IETF network slices {{I-D.ietf-teas-ietf-network-slices}}.
 For example, this can apply to security-related target properties and path selection, such as allowing or disallowing sending flows over paths that involve specific networks or nodes to enforce traffic policies or mandating that all enterprise traffic goes through a specific firewall.
 
-Note, there may be trade-offs between path properties (e.g., One-Way Delay and Link Capacity), and entities may influence these trade-offs with their choices.
+Care needs to be taken when selecting paths based on observed path properties, as path properties that were previously measured may not be helpful in predicting current or future path properties and such path selection may lead to unintended feedback loops. Also, there may be trade-offs between path properties (e.g., One-Way Delay and Link Capacity), and entities may influence these trade-offs with their choices.
+
 As a baseline, a path selection algorithm should aim to not perform worse than the default case most of the time.
 
 Path selection can be done either by the communicating node(s) or by other entities within the network:
--- a/draft-irtf-panrg-path-properties.md
+++ b/draft-irtf-panrg-path-properties.md
@@ -39,6 +39,8 @@ informative:
 
     I-D.ietf-alto-performance-metrics:
 
+    I-D.ietf-teas-ietf-network-slices:
+
     RFC1930:
 
     RFC2616:
@@ -136,6 +138,10 @@ Assessed property:
 : An approximate calculation or assessment of the value of a property. An assessed property includes the reliability of the calculation or assessment. The notion of reliability depends on the property.
 For example, a path property based on an approximate calculation may describe the expected median one-way latency of packets sent on a path within the next second, including the confidence level and interval. A non-numerical assessment may instead include the likelihood that the property holds.
 
+Target property:
+: An objective that is set for a property over a path element, subpath, or path.
+Note that a target property can be set for observable properties, such as one-way delay, but also for properties that are not (easily) observable, such as inclusion of certain nodes on a path.
+
 ## Terminology usage for specific technologies
 
 The terminology defined in this document is intended to be general and applicable to existing and future path-aware technologies.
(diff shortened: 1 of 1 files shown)


draft-irtf-panrg-path-properties.md
```diff
@@ -136,6 +138,10 @@ Assessed property:
 : An approximate calculation or assessment of the value of a property. An assessed property includes the reliability of the calculation or assessment. The notion of reliability depends on the property.
 For example, a path property based on an approximate calculation may describe the expected median one-way latency of packets sent on a path within the next second, including the confidence level and interval. A non-numerical assessment may instead include the likelihood that the property holds.
 
+Target property:
+: An objective that is set for a property over a path element, subpath, or path.
+Note that a target property can be set for observable properties, such as one-way delay, but also for properties that are not (easily) observable, such as inclusion of certain nodes on a path.
```
theri, line 143: I wonder if this section should say "observed" rather than "observable" properties, as that's the defined term, and I think it's obvious enough that OWD being an "observed" property just means that it can be observed, given appropriate measurement tools etc.
Then, the second part could become "properties that cannot be observed by the entity setting the target" - surely, *somebody* could observe whether the node is on the path, but the entity setting the target may not have that insight because the node is not in their domain. (No need to get that specific in the definition, I just wanted to illustrate how I think about it - "inclusion of certain nodes on a path" is not inherently non-observable, it just might be non-observable to a certain entity, in analogy to the SLE in [Section 4.1 of draft-ietf-teas-ietf-network-slices](https://www.ietf.org/archive/id/draft-ietf-teas-ietf-network-slices-14.html#name-objectives-for-ietf-network))

draft-irtf-panrg-path-properties.md
```diff
@@ -158,14 +164,18 @@ Therefore, a new technology may implement an existing use case related to differ
...
-As an example of performance-related path selection,
-an entity may prefer paths with performance properties that best match application requirements.
-For example, for sending a small delay sensitive query, the entity may select a path with a short One-Way Delay,
-while for retrieving a large file, it may select a path with high Link Capacities on all links.
+Care needs to be taken when selecting paths based on observed path properties, as path properties that were previously measured may not be helpful in predicting current or future path properties and such path selection may lead to unintended feedback loops.
+
+Entities may select their paths to fulfill a specific goal by specifying target properties for their selected paths, e.g., related to performance or security.
+
+Target properties relating to network performance typically refer to observable properties, such as delay, loss, or bandwidth.
```
theri, line 171: If we change "observable" to "observed" above, we should do the same here.

Also, let's use examples that are actually defined below in Section 4, such as One-Way Delay, One-Way Packet Loss, and Link Capacity.

draft-irtf-panrg-path-properties.md
```diff
@@ -158,14 +164,18 @@ Therefore, a new technology may implement an existing use case related to differ
...
+Entities then select paths based on their target property and the assessed property of the available paths that best match the application requirements.
+For such performance-related target properties, the observed property is similar to a service level indicator (SLI) and the assessed property is similar to a service level objective (SLO) for IETF network slices {{I-D.ietf-teas-ietf-network-slices}}.
+For example, for sending a small delay-sensitive query, the entity selects a path with a short One-Way Delay, while for retrieving a large file, it selects a path with high Link Capacities on all links.
```
theri, line 174:
```suggestion
As an example path selection strategy, for sending a small delay-sensitive query, an entity may select a path with a short One-Way Delay, while for retrieving a large file, it may select a path with high Link Capacities on all links.
```

Suggest to tweak the phrasing here to make sure to not imply that this is the only possible path selection strategy here.

draft-irtf-panrg-path-properties.md
```diff
@@ -158,14 +164,18 @@ Therefore, a new technology may implement an existing use case related to differ
...
+
+Target properties relating to security are typically not (directly) observable, similar to service level expectations (SLEs).
+Examples are allowing or disallowing sending flows over paths that involve specific networks or nodes to enforce traffic policies or mandating that all enterprise traffic goes through a specific firewall.
```
theri, line 177: Is this really true as a general statement? I think it'd depend on the specific technology and/or context.
I suggest not making this general statement and instead simply talking about properties that cannot be directly observed. I think this might also possibly occur for performance-related path properties, e.g., some entity is unable to measure due to lack of vantage point in the network where it matters, due to encryption, etc.
I'd like to err on the side of being less prescriptive and general here.

So, for example, this could be something like:

```suggestion
It is also possible for an entity to set target properties which it cannot (directly) observe, similar to service level expectations (SLEs) for IETF network slices {{I-D.ietf-teas-ietf-network-slices}}.
For example, this can apply to security-related target properties and path selection, such as allowing or disallowing sending flows over paths that involve specific networks or nodes to enforce traffic policies or mandating that all enterprise traffic goes through a specific firewall.
```

theri (commented):
Thank you for taking a stab at this PR!
I think this is going in the right direction, but I have some fine-tuning comments.

draft-irtf-panrg-path-properties.md
cyrill-k: I like both suggestions.
Also, discussing that some entity may be able to observe a property while another entity cannot is more in line with our definition of a path (i.e., partial visibility and different levels of abstraction).

What do you think about the following change?

```suggestion
Note that a target property can be set for observed properties, such as one-way delay, but also for properties that cannot be observed by the entity setting the target, such as inclusion of certain nodes on a path.
```

draft-irtf-panrg-path-properties.md
cyrill-k: Agreed, let's also change observable to observed here.
I was thinking to use more general terms here since we are still explaining the terminology and these specific property terms are not yet introduced, but I think they are quite self-explanatory so I agree it's probably better to use the actual property names.

How about the following change?
```suggestion
Target properties relating to network performance typically refer to observed properties, such as One-Way Delay, One-Way Packet Loss, and Link Capacity.
```

draft-irtf-panrg-path-properties.md
cyrill-k: Good point, I agree.

draft-irtf-panrg-path-properties.md
cyrill-k: I agree we should clarify that whether a property is observable or not depends on the _specific entity_ that observes the property.

cyrill-k:
thanks for the textual improvements!
I committed your suggestions and added two suggestions based on your feedback.
If the suggestions look good for you, please add them and merge the PR.

draft-irtf-panrg-path-properties.md
theri: This looks good to me, thank you!

draft-irtf-panrg-path-properties.md
theri: Yep, looks good, thanks.

theri:
> Thanks for the textual improvements! I committed your suggestions and added two suggestions based on your feedback.
> 
> Regarding the merge of texts relating to the usefulness of path properties discussed in #75, we could merge the paragraph talking about the property trade-offs ("Note, there may be trade-offs between path properties ...") with the first paragraph: "... may lead to unintended feedback loops. **Also,** there may be trade-offs between path properties ..." What do you think?

I agree that these should be merged.

To me, it would make most sense to move the text in the first paragraph (starting with "Care needs to be taken […]" and ending with "[…] may lead to unintended feedback loops.") a bit down and then embed it there.
That way, we'd first discuss the general concepts around objectives and target properties, and then have the caveats all in one place later.
I will make a suggestion.

draft-irtf-panrg-path-properties.md
```diff
@@ -158,14 +164,18 @@ Therefore, a new technology may implement an existing use case related to differ
...
+As an example path selection strategy, for sending a small delay-sensitive query, an entity may select a path with a short One-Way Delay, while for retrieving a large file, it may select a path with high Link Capacities on all links.
+
+It is also possible for an entity to set target properties which it cannot (directly) observe, similar to service level expectations (SLEs) for IETF network slices {{I-D.ietf-teas-ietf-network-slices}}.
+For example, this can apply to security-related target properties and path selection, such as allowing or disallowing sending flows over paths that involve specific networks or nodes to enforce traffic policies or mandating that all enterprise traffic goes through a specific firewall.
+
 Note, there may be trade-offs between path properties (e.g., One-Way Delay and Link Capacity), and entities may influence these trade-offs with their choices.
 As a baseline, a path selection algorithm should aim to not perform worse than the default case most of the time.
```
theri, line 180:
```suggestion
Entities may select their paths to fulfill a specific goal by specifying target properties for their selected paths, e.g., related to performance or security.

Target properties relating to network performance typically refer to observed properties, such as One-Way Delay, One-Way Packet Loss, and Link Capacity.
Entities then select paths based on their target property and the assessed property of the available paths that best match the application requirements.
For such performance-related target properties, the observed property is similar to a service level indicator (SLI) and the assessed property is similar to a service level objective (SLO) for IETF network slices {{I-D.ietf-teas-ietf-network-slices}}.
As an example path selection strategy, for sending a small delay-sensitive query, an entity may select a path with a short One-Way Delay, while for retrieving a large file, it may select a path with high Link Capacities on all links.

It is also possible for an entity to set target properties which it cannot (directly) observe, similar to service level expectations (SLEs) for IETF network slices {{I-D.ietf-teas-ietf-network-slices}}.
For example, this can apply to security-related target properties and path selection, such as allowing or disallowing sending flows over paths that involve specific networks or nodes to enforce traffic policies or mandating that all enterprise traffic goes through a specific firewall.

Care needs to be taken when selecting paths based on observed path properties, as path properties that were previously measured may not be helpful in predicting current or future path properties and such path selection may lead to unintended feedback loops. Also, there may be trade-offs between path properties (e.g., One-Way Delay and Link Capacity), and entities may influence these trade-offs with their choices.

As a baseline, a path selection algorithm should aim to not perform worse than the default case most of the time.
```

theri: I was hoping for better rendering.
I really just took the first paragraph of the marked text and moved it to almost the bottom, and then changed a "Note" to an "Also", and added a newline before the last sentence because it's a slightly separate point and a very important one.

What do you think?

cyrill-k: That looks good :+1:

PR #82, opened 2022-12-15
Add two statements described in RFC 5743

cyrill-k:
Address Colin's points mentioned in #78


theri (approved):
Thanks for adding the statements, Cyrill! This looks good to me.

Issue #79, opened 2022-12-13
(Colin's Review): Clarification in Section 3.1


boucadair:
In Section 3.1, 
`“Entities may select their paths to fulfill a specific goal by specifying target properties for their selected paths” `

Is the intent that an entity will specify a target property and the network will select a path to fulfil that property, or that the entity will somehow measure the property and choose a path that meets it?

cyrill-k:
I would say that both are examples for a use case related to path selection.
We wanted to express that target properties describe an entity's **intent** for which path properties are desirable for their paths.

In the last paragraph of Section 3.1, we discuss this point in more details: "Path selection can be done either by the communicating node(s) or by other entities within the network: A network (e.g., an AS) can adjust
its path selection for internal or external routing based on path properties. ..."

Maybe we can clarify the sentence in question to: "Entities can achieve a specific goal by specifying target properties for their paths. Such paths can be provided by the network or selected by the entity itself from all available paths."

csperkins:
That rewording seems clearer.

PR #83, opened 2022-12-15
clarify that paths can be selected by the network or the endpoints

cyrill-k:
This clarifies that path selection can be done by the network or the entity requesting a certain target property (see #79)

### DIFF
--- a/draft-irtf-panrg-path-properties.md
+++ b/draft-irtf-panrg-path-properties.md
@@ -181,8 +181,8 @@ Therefore, a new technology may implement an existing use case related to differ
 
 Nodes may be able to send flows via one (or a subset) out of multiple possible paths, and an entity may be able to influence the decision which path(s) to use.
 Path Selection may be feasible if there are several paths to the same destination (e.g., in case of a mobile device with two wireless interfaces, both providing a path), or if there are several destinations, and thus several paths, providing the same service (e.g., Application-Layer Traffic Optimization (ALTO) {{RFC5693}}, an application layer peer-to-peer protocol allowing endpoints a better-than-random peer selection).
-Entities can achieve a specific goal by specifying target properties for their paths, e.g., related to performance or security.
-Such paths can be provided by the network or selected by the entity itself from all available paths.
+Entities can express their intent to achieve a specific goal by specifying target properties for their paths, e.g., related to performance or security.
+Then, paths can be selected which best meet the target properties, e.g., the entity can select these paths from all available paths or express the target properties to the network and rely on the network to select appropriate paths.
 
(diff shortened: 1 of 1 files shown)


draft-irtf-panrg-path-properties.md
```diff
@@ -181,7 +181,8 @@ Therefore, a new technology may implement an existing use case related to differ
 
 Nodes may be able to send flows via one (or a subset) out of multiple possible paths, and an entity may be able to influence the decision which path(s) to use.
 Path Selection may be feasible if there are several paths to the same destination (e.g., in case of a mobile device with two wireless interfaces, both providing a path), or if there are several destinations, and thus several paths, providing the same service (e.g., Application-Layer Traffic Optimization (ALTO) {{RFC5693}}, an application layer peer-to-peer protocol allowing endpoints a better-than-random peer selection).
-Entities may select their paths to fulfill a specific goal by specifying target properties for their selected paths, e.g., related to performance or security.
+Entities can achieve a specific goal by specifying target properties for their paths, e.g., related to performance or security.
+Such paths can be provided by the network or selected by the entity itself from all available paths.
```
theri, line 185:
```suggestion
Entities can express their intent to achieve a specific goal by specifying target properties for their paths, e.g., related to performance or security.
Then, a path can be selected which best meets the target property, e.g., the entity can select this path from all available paths or the entity can express the target property to the network and rely on the network to select an appropriate path.
```

I agree that the main point here is that the target property describes an intent, and the implementation of the path selection is left open.
I rephrased this change to express that _somebody_ is going to select from the available paths (and not necessarily reconfigure a single existing path to meet a target), as this is the "Path selection" section.

theri (commented):
Thanks for proposing text!
I suggest a subtle change in the phrasing. I'm sure there's an easier way to express this, but I didn't find it.

draft-irtf-panrg-path-properties.md
cyrill-k: I like your changes. The only thing that I would change is use the plural instead of singular for path since path selection might also be used in a multipath setting. Also I shortened the second sentence a bit.
```suggestion
Entities can express their intent to achieve a specific goal by specifying target properties for their paths, e.g., related to performance or security.
Then, paths can be selected which best meet the target property, e.g., the entity can select these paths from all available paths or express the target property to the network and rely on the network to select appropriate paths.
```

boucadair: I would also use "target properties" instead of "target property" in the second sentence.

theri: Thanks to both of you! I'll apply the suggested changes.

PR #85, opened 2023-01-09
Update acknowledgments

theri:
Moved most reviewer names to an alphabetical list sorted by last name, added Colin.


cyrill-k:
I removed a spurious "and" and I removed the "/or" since it seems clear to me that not all mentioned people did all of the three tasks.

Otherwise it looks good to me :+1:

PR #96, opened 2023-02-14
Clarify that the path selection should be better than not selecting a path

### LINKED ISSUE #87: IRSG Review (Dave Oran): Perform no worse than the default case

> In 3.1 Path Selection you say “…a path selection algorithm should aim to not perform worse than the default case most of the time.” I’m not entirely sure what constitutes the “default case”. Is it the case where no path selection is made by the user, or where no path selection is possible? If the former, I can certainly imagine someone intentionally selecting a worse performing path for background traffic. Perhaps better wording would be “…a path selection algorithm should aim to do a better job of accommodating the user’s requirements than the default case of not selecting a path”.

theri:
Fix #87 - The default case is that we don't do path selection at all. 
What exactly happens then will depend on the context, but I'd say this usually means that a single "default" path is used, which may be statically configured. But I don't think it's necessary to spell this out here.
I am pretty sure that we don't want to compare it to a random selection, because I've often seen cases where the default path is "good" and the alternative path is just unsuitable. In such a case, if you select the path randomly, you'll perform much worse than the default of not doing path selection at all.

### DIFF
--- a/draft-irtf-panrg-path-properties.md
+++ b/draft-irtf-panrg-path-properties.md
@@ -201,7 +201,7 @@ Care needs to be taken when selecting paths based on observed path properties, a
 Finally, path selection may impact fairness.
 For example, if multiple entities concurrently attempt to meet their target properties using the same network resources, one entity's choices may influence the conditions on the path as experienced by flows of another entity.
 
-As a baseline, a path selection algorithm should aim to not perform worse than the default case of not selecting a path most of the time.
+As a baseline, a path selection algorithm should aim to do a better job at meeting the target properties or accomodating the user's requirements than the default case of not selecting a path most of the time.
 
 Path selection can be done either by the communicating node(s) or by other entities within the network:
 A network (e.g., an AS) can adjust its path selection for internal or external routing based on path properties.
--- a/draft-irtf-panrg-path-properties.md
+++ b/draft-irtf-panrg-path-properties.md
@@ -201,7 +201,7 @@ Care needs to be taken when selecting paths based on observed path properties, a
 Finally, path selection may impact fairness.
 For example, if multiple entities concurrently attempt to meet their target properties using the same network resources, one entity's choices may influence the conditions on the path as experienced by flows of another entity.
 
-As a baseline, a path selection algorithm should aim to do a better job at meeting the target properties or accomodating the user's requirements than the default case of not selecting a path most of the time.
+As a baseline, a path selection algorithm should aim to do a better job at meeting the target properties, and consequently accommodating the user's requirements, than the default case of not selecting a path most of the time.
 
 Path selection can be done either by the communicating node(s) or by other entities within the network:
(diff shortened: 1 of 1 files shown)


draft-irtf-panrg-path-properties.md
```diff
@@ -201,7 +201,7 @@ Care needs to be taken when selecting paths based on observed path properties, a
 Finally, path selection may impact fairness.
 For example, if multiple entities concurrently attempt to meet their target properties using the same network resources, one entity's choices may influence the conditions on the path as experienced by flows of another entity.
 
-As a baseline, a path selection algorithm should aim to not perform worse than the default case most of the time.
+As a baseline, a path selection algorithm should aim to not perform worse than the default case of not selecting a path most of the time.
```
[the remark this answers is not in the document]
theri: Yes, I think what we have is sufficient. I could be convinced to reevaluate if other reviewers have a strong opinion here.

theri:
As with the other PR, I added Med and Dave in case they want to chime in as well.
If I don't hear anything within a week (by March 3rd, 9am PT) I will merge.

daveoran:
I'm ok with this, but I'll try one last time to suggest that "perform better"  doesn't capture the essence of why a particular path might be selected by a user. As I tried (i guess unsuccessfully) to get across, a user might intentionally select a path that performs worse (in terms of the usual "performance" measures like bandwidth, latency, etc.) in order to satisfy some other objective function they have, like lowering cost, or increasing the probability that some other path selection request will succeed for other traffic they generate. That's why i suggested the wording "…a path selection algorithm should aim to do a better job of accommodating the user’s requirements than the default case of not selecting a path”.

theri:
Ohh, I got you now, thank you for reiterating this aspect of your comment.
I agree that the objective is not always related to performance, and I changed the text to reflect that. 

Everyone, please let me know if that sounds good to you.

daveoran (approved):

draft-irtf-panrg-path-properties.md
```diff
@@ -201,7 +201,7 @@ Care needs to be taken when selecting paths based on observed path properties, a
 Finally, path selection may impact fairness.
 For example, if multiple entities concurrently attempt to meet their target properties using the same network resources, one entity's choices may influence the conditions on the path as experienced by flows of another entity.
 
-As a baseline, a path selection algorithm should aim to not perform worse than the default case most of the time.
+As a baseline, a path selection algorithm should aim to do a better job at meeting the target properties or accomodating the user's requirements than the default case of not selecting a path most of the time.
```
boucadair, line 204:
```suggestion
As a baseline, a path selection algorithm should aim to do a better job at meeting the target properties, and consequently accommodating the user's requirements, than the default case of not selecting a path most of the time.
```

cyrill-k: I like that phrasing better as well

cyrill-k (commented):
This looks good to me.

cyrill-k (approved):

PR #98, opened 2023-02-14
Editorial fixes

### LINKED ISSUE #94: IRSG Review (Dave Oran): Editorial issues

>  The Abstract says the “document specifies several path properties…”. I wonder if the descriptions of these in the main text in fact constitute a rigorous definition intended to be normative in other documents (i.e. ‘specified’) or informative only, in which case I’d suggest changing the wording to “document identifies several path properties…”

>  First sentence of introduction, suggest changing the wording to “…does not explicitly support endpoint discovery of forwarding paths through the network nor the discovery…”

>  For the first-time reader, it might be worth emphasizing in the introduction that the statement “the document lists several path properties” is not intended to be either exhaustive or definitive.

> The last sentence of the definition of Path says “…these paths do not have to be disjoint.” A forward reference to your definition of Disjointness might be useful here.

> In section 4, third paragraph. you say “…by aggregating them appropriately or dampening their changes to avoiding oscillation.” I don’t think you exactly mean this as you can’t actually damp the rate of change of the actual properties. Instead perhaps say “by aggregating them appropriately or applying a dampening function to their changes to avoid oscillation.”

> A bit further down you say “..a NAT typically stays on a specific path”. Perhaps “a NAT typically stays on a particular path” since the path may not be “specific”.

> This is probably unfair to a large community, but my view is that in the list of “access technologies “2G, 3G, 4G, or 5G” don’t really belong since those are basically marketing labels and not technologies. Can we get away with simply saying e.g., 3GPP cellular, or 802.11 WiFi”?

> A bit further down you have the term “cellular backbone network”. I may be wrong, but I think that community instead uses the term “cellular core network”

> In “Service Functions” you list “TCP Optimizers”. Might “TCP Performance Enhancing Proxies” be more appropriate for this context?

> For “Administrative Domain” you say “…individual or an organization that owns a path element…” - I’m not sure ownership is the defining characteristic rather than perhaps “…individual or an organization that controls access to a path element…”

(At first glance, these all seem like editorial fixes that won't require much discussion. If you disagree that this is true for any of the above comments, please let me know.)

theri:
I still think these are all editorial. Fixes #94

### DIFF
--- a/draft-irtf-panrg-path-properties.md
+++ b/draft-irtf-panrg-path-properties.md
@@ -83,21 +83,21 @@ This document is a product of the Path Aware Networking Research Group (PANRG).
 Path properties express information about paths across a network and the services provided via such paths.
 In a path-aware network, path properties may be fully or partially available to entities such as endpoints.
 This document defines and categorizes path properties.
-Furthermore, the document specifies several path properties which might be useful to endpoints or other entities,
+Furthermore, the document identifies several path properties which might be useful to endpoints or other entities,
 e.g., for selecting between paths or for invoking some of the provided services.
 
 --- middle
 
 # Introduction
 
-The current Internet architecture does not explicitly support endpoint discovery of forwarding paths through the network as well as the discovery of properties and services associated with these paths.
+The current Internet architecture does not explicitly support endpoint discovery of forwarding paths through the network nor the discovery of properties and services associated with these paths.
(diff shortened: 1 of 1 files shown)


cyrill-k:
All of these changes look good to me.

theri:
No pushback on the "editorial" part, so I'm merging this now.

Issue #91, opened 2023-02-07
IRSG Review (Dave Oran): Path MTU only applies to IP?


theri:
From the IRSG review by Dave Oran:

> In “Path MTU”, do you intend this to apply only to “an IP packet that can be transmitted”, or any packet sent over that path? Also, it might be clearer to say “…can be transmitted over the entire path without fragmentation”.

theri:
I think we apply MTU only to IP here because we were looking at Path MTU Discovery [RFC1191](https://datatracker.ietf.org/doc/rfc1191/) in this context, which is legacy now.
And we defined many other path properties only in relation to IP, such as several performance path properties further down.

That said, we don't cite RFC 1191 or any other document which defines MTU, and I'm not sure if we care much either way.

Does anyone have any strong opinions here? If not, I would lean towards removing the "IP" here and leave it open to the definition of "packet", which we never specify as strictly IP.

PR #100, opened 2023-02-27
Change "IP MTU" to just "MTU" in the definition

theri:
Based on Dave Oran's review comment #91, can we just remove "IP" here to leave it as just "MTU"?
(It turns out that the draft even talks about an "Ethernet MTU" in Section 4...)

### DIFF
--- a/draft-irtf-panrg-path-properties.md
+++ b/draft-irtf-panrg-path-properties.md
@@ -272,7 +272,7 @@ Symmetric Path:
 For example, a path which consists of layer 3 switches and links between them and a reverse path with the same path elements but in reverse order are considered "routing" symmetric, as the same path elements on the same level of abstraction (IP forwarding) are traversed in the opposite direction.
 
 Path MTU:
-: The maximum size, in octets, of an IP packet that can be transmitted without fragmentation.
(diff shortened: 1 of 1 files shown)


draft-irtf-panrg-path-properties.md
```diff
@@ -272,7 +272,7 @@ Symmetric Path:
 For example, a path which consists of layer 3 switches and links between them and a reverse path with the same path elements but in reverse order are considered "routing" symmetric, as the same path elements on the same level of abstraction (IP forwarding) are traversed in the opposite direction.
 
 Path MTU:
-: The maximum size, in octets, of an IP packet that can be transmitted without fragmentation.
+: The maximum size, in octets, of a packet that can be transmitted without fragmentation.
```
boucadair, line 275:
```suggestion
: The maximum size, in octets, of a packet (or frame) that can be transmitted without fragmentation.
```

theri: Thanks for the suggestion, I'll use it but without the brackets :slightly_smiling_face:

cyrill-k (approved):
Looks good to me

PR #102, opened 2023-02-27
Add a note about levels of abstraction to Disjointness definition

### LINKED ISSUE #89: IRSG Review (Dave Oran): Disjointness

> For “Disjointness” I get what you have as the definition, however there is some recursive nature to this when path elements can be composed through multiple layers. A particular path element might be disjoint from another at layer N but in fact are not at layer N-k. What the user probably cares about is the degree of fate sharing between two path elements, which is not entirely captured in the description (but possibly might be by careful application of the Jaccard coefficient projects across layers). Probably not much can be said to go into this without rat-holing though.

theri:
Resolve #89


cyrill-k (approved):
I like this additional sentence to clarify how disjointness interacts with the different levels of abstraction.

PR #103, opened 2023-02-27
Clarify symmetric paths and abstraction levels

### LINKED ISSUE #90: IRSG Review (Dave Oran): Symmetric paths

> In “Symmetric path” do all the elements need to be symmetric or only the nodes? Do we care about packets traversing LAGs or or other parallel connections between a pair of nodes constituting different paths? Just a question, really…

theri:
Resolve #90

### DIFF
--- a/draft-irtf-panrg-path-properties.md
+++ b/draft-irtf-panrg-path-properties.md
@@ -270,6 +270,7 @@ Disjointness:
 Symmetric Path:
 : Two paths are symmetric if the path and its reverse path consist of the same path elements on the same level of abstraction, but in reverse order.
 For example, a path which consists of layer 3 switches and links between them and a reverse path with the same path elements but in reverse order are considered "routing" symmetric, as the same path elements on the same level of abstraction (IP forwarding) are traversed in the opposite direction.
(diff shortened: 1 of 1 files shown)


daveoran (approved):

draft-irtf-panrg-path-properties.md
```diff
@@ -270,6 +270,7 @@ Disjointness:
 Symmetric Path:
 : Two paths are symmetric if the path and its reverse path consist of the same path elements on the same level of abstraction, but in reverse order.
 For example, a path which consists of layer 3 switches and links between them and a reverse path with the same path elements but in reverse order are considered "routing" symmetric, as the same path elements on the same level of abstraction (IP forwarding) are traversed in the opposite direction.
+Symmetry can depend on the level of abstraction on which the path is defined or modeled: If there are two parallel physical links between two nodes, modeling these links as separately may result in a flow using asymmetric paths, and modeling them as a single virtual link may result in symmetric paths, e.g., if the difference between the two physical links is irrelevant in a particular context.
```
cyrill-k, line 273: I suggest a slight rewording:

```suggestion
Symmetry can depend on the level of abstraction on which the path is defined or modeled: If there are two parallel physical links between two nodes, modeling them as separate links may result in a flow using asymmetric paths, and modeling them as a single virtual link may result in symmetric paths, e.g., if the difference between the two physical links is irrelevant in a particular context.
```

cyrill-k (approved):

